PR #1, opened 2016-09-28
relaxed int conditions on subdomain_id

alsgregory:
Allow `numpy.integer` types to be accepted as well as `str` and `int` in `subdomain_id`


dham:
Makes sense to me. Duck typing and all that.

garth-wells:
UFL is developed at https://bitbucket.org/fenics-project/ufl - this is just a mirror.

PR #4, opened 2019-06-21
L2pullbacks

celdred:
This adds proper L2 pullbacks (ie integral-preserving, designed for n-forms), by defining new elements at the UFL level and introducing an "L2 Piola" mapping. These elements then share the same FIAT/FiNAT classes as the original elements, since from the persRpective of the tabulator the pullback doesn't matter.

Requires https://github.com/firedrakeproject/tsfc/pull/194 and https://github.com/firedrakeproject/firedrake/pull/1463


dham:
FEniCS haven't yet fully moved UFL development to GitHub yet, so you need to file this PR on Bitbucket.

wence-:
Opened upstream here https://bitbucket.org/fenics-project/ufl/pull-requests/114/l2pullbacks/diff

wence-:
In master.

wence-:
@celdred can you close this?

PR #5, opened 2019-06-28
Mimetic spectral elements

celdred:
This pull request adds mimetic spectral elements (described at https://www.sciencedirect.com/science/article/pii/S0021999113006414 and https://arxiv.org/abs/1111.4304), both the primal and dual complexes. This is done by adding the relevant 1D H1 element for the dual complex: Extended Gauss Legendre (EGL); and the L2 edge elements associated with GLL and EGL. These are non-Ciarlet finite elements that use a basis that histopolates (http://people.math.sfu.ca/~nrobidou/public_html/prints/histogram/histogram.pdf) rather than interpolates. The nD deRham complex on hypercubes is then accessible using the variant keyword (either “mse” or “dualmse”) with the relevant elements (DQ, Q, RTCF, etc.).

It also requires correct (integral-preserving) L2 pullbacks from https://github.com/FEniCS/ufl/pull/4.

### DIFF
--- a/test/test_elements.py
+++ b/test/test_elements.py
@@ -199,3 +199,25 @@ def test_radau():
 
         element = FiniteElement("Radau", cell, degree)
         assert element == eval(repr(element))
+
+
+def test_mse():
+    for degree in (2, 3, 4, 5):
+        element = FiniteElement('EGL', interval, degree)
+        assert element == eval(repr(element))
+
+        element = FiniteElement('EGL-Edge', interval, degree - 1)
+        assert element == eval(repr(element))
+
+        element = FiniteElement('EGL-Edge L2', interval, degree - 1)
+        assert element == eval(repr(element))
+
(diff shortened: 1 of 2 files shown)


miklos1:
Okay, so this pull request is not very useful here. FIAT and UFL pull requests should be filed against upstream to be merged: https://bitbucket.org/fenics-project/ufl

If you want to receive code review from the Firedrake team before filing the pull request upstream, then please file the pull request against `firedrakeproject:master` and not `FEniCS:master` (as is the case with this PR). Since this GitHub mirror is quite out of date, the diff shown for this PR is practically unreadable.

wence-:
I will squash and rebase this is that ufl is being developed here rather than bitbucket.

wence-:
I think this is now ready for review. As @celdred says, it registers a bunch of new elements.

[Aside, should we think of/advocate a way that third party "clients" of UFL can register new elements without having to put them all in here?]

PR #15, opened 2020-01-28
Extend replace to be able to handle non-terminals

dham:
Replace currently only replaces terminals. In fact, there are circumstances where replacing non-terminals is convenient, and this just works for many cases. This PR relaxes the checks in UFL to allow non-terminals to be replaced.

This PR incidentally includes an expanded comment explaining some pre-existing but slightly cryptic code in which replace sometimes applies derivatives.


michalhabera:
Thanks, @dham , I already use your replacer in some custom code. If you can just update docstring for `replace()`. The old "Replace terminal objects in expression." is no longer true.

dham:
> Thanks, @dham , I already use your replacer in some custom code. If you can just update docstring for `replace()`. The old "Replace terminal objects in expression." is no longer true.

Good point. I've just fixed that.

PR #17, opened 2020-03-13
Allow VectorElement to take a variant argument.

pefarrell:


wence-:
Can you do this for `TensorElement` too, then I'll merge.

ufl/finiteelement/mixedelement.py
```diff
@@ -263,6 +263,8 @@ def __init__(self, family, cell=None, degree=None, dim=None,
                viewed as k-form), ignored if family is a FiniteElement
             quad_scheme
                The quadrature scheme (optional), ignored if family is a FiniteElement
+            variant
```
garth-wells, line 266: I think this needs a bit more explanation. I don't understand the docstring.

miklos1: This variant thing was first introduced to distinguish between GL, GLL, and equispaced Langrange elements without a profusion of names: Q, dQ, RTCE, RTCF, NCE, NCF type elements could be built from equispaced Lagrange as well as from spectral ingredients.

Later uses of this include RT and N1curl elements being available with different nodal basis, etc.

miklos1: c.f. FEniCS/fiat#52

miklos1: For further reference, see [this UFL issue](https://bitbucket.org/fenics-project/ufl/issues/92) and [this UFL pull request](https://bitbucket.org/fenics-project/ufl/pull-requests/69).

PR #26, opened 2020-05-18
External operators

nbouziani:
This PR follows the discussion initiated here:  https://github.com/firedrakeproject/ufl/issues/15, cf. the following PR https://github.com/firedrakeproject/ufl/pull/16

`ExternalOperator` is a subclass of `Coefficient` which carries operands and a derivative multi-index. Its purpose is to enable the incorporation in a form of operators that are not directly expressible in UFL. A number of important classes of function fall into this category:

  - Implicit constitutive relations. I.e. nonlinear solids or fluids where the stress is computed by solving a point wise nonlinear equation every time the operator is evaluated.

  - Explicit functions not easily written in UFL such as neural nets which might be used in learned parametrisations or learned regularisers for inverse problems.

  - Parametrisations which are inconvenient to write in UFL because they, for example, contain a lot of switching and interpolation from lookup tables.

`ufl.ExternalOperator` handles the symbolic side (e.g. differentiation), a practical implementation is needed to state how does one correlate the operands to evaluate the `ExternalOperator`.

This practical implementation is already implemented in Firedrake: see firedrakeproject/firedrake#1674

The implementation of the adjoint capability is also already done, enabling to tackle more generally PDE-constrained optimization problems.

Examples of applications cover:

- Implicit constitutive relations (e.g. Glen's flow law in glaciology)
- Regularizer for inverse problems (e.g. seismic inversion)
- Subgrid parametrization

Mathematically, an `ExternalOperator` **N** can be defined as follows:

    V_1 x ... x V_k  --> X
    u1, ..., uk    |--> N(u1, ..., uk)

where `u1, ..., uk` are Coefficients belonging respectively to the function spaces `V_1, ..., V_k`, and where `X` is the function space associated to the external operator **N**.

In practice, when assembling a form containing **N**, we:

   - Interpolate the operands (u1, ..., uk) into X
   - Evaluate N(u1^{X}, ..., uk^{X})

Obviously, we need to explicitly define how to evaluate **N**, this is done at the Firedrake stage where we have an `AbstractPointwiseOperator` class that subclasses `ExternalOperator` and that provides a method to evaluate the `ExternalOperator`.

For instance, for the implicit constitutive law example the evaluate method will solve the nonlinear constitutive law pointwisely.

For the explicit function case, here is toy example illustrating how looks the pseudo code to solve the poisson's equation with a rhs `f` where `inner(u-f, v)` is replaced by `inner(p2, v)`:

    ...
    V = FunctionSpace(...)
    ...
    u = Function(V)
    ...
    p = point_expr(lambda x, y: x - y, function_space=V)
    p2 = p(u, f)
    ...
    F = inner(grad(p2), grad(v) * dx + inner(p2, v) * dx
    solve(F == 0, u, ...)
    ...


meg-simula:
I like this concept! Do not have time to look at the code at the moment, what do you think @michalhabera @wence- ?

dham:
We've had an offline review from @wence- and we think that, while the interface is right, we are wrong to think of this as a terminal. In fact it should be an operator. We think that making this change will make this PR a lot less intrusive in the UFL code by removing a whole load of special casing. @nbouziani will push updates soon.

michalhabera:
Thanks @dham , I like the concept as well as @meg-simula . I am just missing a pure UFL example code, @nbouziani provided `point_expr` which is a black box wrt. UFL. I will have a look at the code after announced updates.

ufl/log.py
```diff
@@ -242,4 +242,4 @@ class UFLValueError(UFLException):
 for foo in log_functions:
     exec("%s = ufl_logger.%s" % (foo, foo))
 
-set_level(DEPRECATE)  # noqa
+set_level(DEPRECATE)  # noqa
```
wence-, line 245: Bad whitespace.

ufl/formoperators.py
```diff
@@ -215,7 +216,7 @@ def _handle_derivative_arguments(form, coefficient, argument):
     for (c, a) in zip(coefficients, arguments):
         if c.ufl_shape != a.ufl_shape:
             error("Coefficient and argument shapes do not match!")
-        if isinstance(c, Coefficient) or isinstance(c, SpatialCoordinate):
+        if isinstance(c, (Coefficient, ExternalOperator)) or isinstance(c, SpatialCoordinate):
```
wence-, line 219: style: prefer `isinstance(c, (Coefficient, ExternalOperator, SpatialCoordinate))`

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        coefficient_signature = self.coefficient._ufl_signature_data_(renumbering)
+        return ("ExternalOperator", self._external_operator_type, *coefficient_signature, *self.derivatives)
+
+    def __eq__(self, other):
+        if not isinstance(other, ExternalOperator):
+            return False
+        if self is other:
+            return True
+        return (self._count == other._count and
+                self._ufl_function_space == other._ufl_function_space)
```
wence-, line 278: Is this basically the same equality that `Coefficient` provides?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        self.coefficient._count = value
+
+    @property
+    def ufl_shape(self):
+        "Return the UFL shape of self.coefficient."
+        return self.coefficient._ufl_shape
+
+    @ufl_shape.setter
+    def ufl_shape(self, shape):
+        self.coefficient._ufl_shape = shape
```
wence-, line 141: Why do you need to be able to set the shape?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+
+    _external_operator_type = ''
+
+    def __init__(self, *operands, function_space, derivatives=None, count=None, coefficient=None, arguments=()):
+        r"""
+        :param operands: operands on which acts the :class:`ExternalOperator`.
+        :param function_space: the :class:`.FunctionSpace`,
+        or :class:`.MixedFunctionSpace` on which to build this :class:`Function`.
+        Alternatively, another :class:`Function` may be passed here and its function space
+        will be used to build this :class:`Function`.  In this case, the function values are copied.
```
wence-, line 39: These lines need to be indented further for the RST to be correct.

wence-: This looks like documentation from Firedrake (UFL has no `Function` object, nor does it have values).

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        :param derivatives: tuple scecifiying the derivative multiindex.
+        :param count: count of the associated coefficient
+        :param coefficient: ufl.Coefficient associated to the ExternalOperator representing what is produced by the operator
```
wence-, line 42: All coefficients come with a count attached, so why require both?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        return ops, args
+
+    def arguments(self):
+        return self._arguments
+
+    def action_args(self):
+        return self._action_args
+
+    @property
+    def _extop_items(self):
```
wence-, line 120: This is a tremendously non-intuitive name.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+                # Therefore, for 'dedu' we need to construct the new TensorElement to end up with the appropriate
+                # function space and we also need to store the VectorElement corresponding to V since it is on this
+                # function_space that we will interpolate the operands.
+                s = ref_coefficient.ufl_shape
+                for i, e in enumerate(self.derivatives):
+                    s += self.ufl_operands[i].ufl_shape * e
+                original_function_space = self._make_function_space(s, sub_element=ref_coefficient.ufl_element())
+                ufl_shape = original_function_space.ufl_element().reference_value_shape()
+            else:
+                ufl_shape = ref_coefficient.ufl_element().reference_value_shape()
```
wence-, line 83: Are you sure you don't mean `value_shape()` ?

ufl/algorithms/compute_form_data.py
```diff
@@ -348,7 +346,31 @@ def compute_form_data(form,
     self.reduced_coefficients = sorted(reduced_coefficients_set,
                                        key=lambda c: c.count())
     self.num_coefficients = len(self.reduced_coefficients)
-    self.original_coefficient_positions = [i for i, c in enumerate(self.original_form.coefficients())
+
+    # Store all the external operators and their derivative multiindex
+    # turning up in the form (Differentiation may have introduced new external operators)
+    extops_pos = {e._extop_master.count(): i for i, e in reversed(list(enumerate(self.original_form.external_operators())))}
```
wence-, line 352: Given you're putting things in a dict, why not just `{e._extop_mast.count(): i for i, e in enumerate(...)}`?

ufl/algorithms/check_arities.py
```diff
@@ -43,6 +43,10 @@ def nonlinear_operator(self, o):
 
     expr = nonlinear_operator
 
+    def external_operator(self, o):
+        res = tuple(map_expr_dag(self, arg, compress=False)[0] for arg, _ in o.arguments())
```
wence-, line 47: Can just do `res = tuple(map(self, (a for a, _ in o.arguments()))` I think. There's no need to invoke the map_expr_dag machinery because the arguments are terminal objects so there's an immediate non-recursive handler for them.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @property
+    def _original_ufl_function_space(self):
+        "Get the function space of this coefficient."
+        return self.coefficient._ufl_function_space
+
+    @_original_ufl_function_space.setter
+    def _original_ufl_function_space(self, func_space):
+        "Get the function space of this coefficient."
+        self.coefficient._ufl_function_space = func_space
```
wence-, line 154: In place modification of things is bad. Especially because now `ufl_function_space` and `original_ufl_function_space` can end up going out of sync. Why do you need this?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        elif not isinstance(coefficient, Coefficient):
+            raise TypeError('Expecting a Coefficient and not %s', type(coefficient))
+        self._coefficient = coefficient
+        self.ufl_shape = ufl_shape
+        self._original_function_space = self.coefficient.ufl_function_space()
+
+        self._extop_dependencies = [self, ]
+
+        if self.derivatives == (0,) * len(self.ufl_operands):
+            self._extop_master = self
```
wence-, line 101: UFL objects are traditionally a DAG, this puts cycles into the expressions, which is bad. Why do you need to track dependencies like this?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,278 @@
...
+        else:
+            reconstruct_op._extop_master = self._extop_master
+            reconstruct_op._extop_dependencies = [e._ufl_expr_reconstruct_(*operands) for e in self._extop_dependencies[1:]]
+        return reconstruct_op
+
+    def _add_dependencies(self, derivatives, args):
+        v = list(self._ufl_expr_reconstruct_(*self.ufl_operands, derivatives=d, arguments=a)
+                 for d, a in zip(derivatives, args))
+        self._extop_master._extop_dependencies.extend(x for x in v if x not in self._extop_master._extop_dependencies)
+        return self
```
wence-, line 246: This dependency stuff smells very "I want to compute something". Does it belong in the symbolics?

ufl/core/external_operator.py
nbouziani: Yes it is, do you think I should use the coefficient produced by the external operator for this? Something like:

```
def __eq__(self, other):
	if not isinstance(other, ExternalOperator):
	            return False
	return self.coefficient.__eq__(other.coefficient)
```

ufl/formoperators.py
nbouziani: Thanks

ufl/log.py
nbouziani: done

doc/sphinx/source/manual/form_language.rst
```diff
@@ -443,6 +443,49 @@ on each, illustrated here::
...
+
+    dpdu = diff(p, u)
+
+is equivalent after processing the expression to::
+
+    dpdu = - sin(u) * p2
+
+where::
+
+    p2 = p._ufl_expr_reconstruct_(cos(u), g, derivatives=(1, 0))
```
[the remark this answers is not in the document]
nbouziani: I have added some explanations, let me know if you think that’s ok.

doc/sphinx/source/manual/form_language.rst
```diff
@@ -443,6 +443,49 @@ on each, illustrated here::
...
+      c = Constant(Cell)
+      p = ExternalOperator(u, g, c, function_space=V)
+
+``ExternalOperator`` has a a symbolic differentiation mechanism, therfore::
+
+    u = Coefficient(V)
+    g = Coefficient(V)
+    p = ExternalOperator(cos(u), g, function_space=V)
+
+    dpdu = diff(p, u)
```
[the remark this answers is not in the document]
nbouziani: done

doc/sphinx/source/manual/form_language.rst
```diff
@@ -443,6 +443,49 @@ on each, illustrated here::
...
+
+      V  = FiniteElement("Lagrange", "triangle", 1)
+      DG0 = FiniteElement("Discontinuous Lagrange", cell, 0)
+
+      u = Coefficient(V)
+      g = Coefficient(DG0)
+      c = Constant(Cell)
+      p = ExternalOperator(u, g, c, function_space=V)
+
+``ExternalOperator`` has a a symbolic differentiation mechanism, therfore::
```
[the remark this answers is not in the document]
nbouziani: done

doc/sphinx/source/manual/form_language.rst
```diff
@@ -443,6 +443,49 @@ on each, illustrated here::
   v1 = Coefficient(DG0v)
   t1 = Coefficient(DG0t)
 
+
+External Operators
+---------------------
+The data type ``ExternalOperator`` subclasses ``Operator``,  it is an operator that takes operands and produces a ``Coefficient`` out of it. In addition, it is equipped with a derivative multi-index. The way to correlate the operands to evaluate the ``ExternalOperator`` is defined at a later stage. Its operands can be ``Coefficient``, ``Expr`` and all their subclasses. As for ``Operators``, their constructors should take these operands as the position arguments, it has its own implementation of ``reconstruct`` which takes into account the operands as well as the derivative multi-index.
+
```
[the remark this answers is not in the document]
nbouziani: Yes, this bit is redundant. I have corrected it. I have also tried to exhibit the composition property, i.e. have other external operators in the operands.

ufl/core/external_operator.py
nbouziani: That’s a remains from the former implementation where ExternalOperator was a Coefficient whose shape was stemming from `original_function_space` and not `function_space`. I removed it, thanks!

ufl/core/external_operator.py
nbouziani: Fixed, thanks!

ufl/core/external_operator.py
nbouziani: This is not needed anymore, I removed it.

ufl/core/external_operator.py
nbouziani: Yes indeed, I was not really imaginative when I wrote this. I had to remove this anyway as it only appears once (in `ufl/algorithms/analysis`)

ufl/algorithms/compute_form_data.py
nbouziani: Because otherwise we end up with the wrong position (it is not a one-to-one relation). Let’s say that we start from an external operator `p(u)` and its derivative `dpdu(u)`, because external operators are sorted respectively with their count value `form.external_operators()` will be [`p`, `dpdu`] as `dpdu` was made after `p` (dpdu.count() > p.count()).  Now to refer to the derivatives of the external operator `p` we store the position of `p`, so `enumerate(...)` does not do the job as it will consider the position of `dpdu` as `dpdu._extop_master = p` and `p._extop_master = p`.

ufl/algorithms/check_arities.py
nbouziani: Actually `arg` here is not necessarily a `ufl.Argument` it can be an expression containing an argument as explained in the class definition. Let’s consider the Gateaux derivative `dp’(u)[v] = dp * v` where `p` and `dp` are ExternalOperators and `dp` has a derivatives multi-index equal to (1,). Currently in the symbolic where the ExternalOperator is global and you want its action you don’t return  `dpdu * v` but `dp[(u,); v]` where `dpdu` is an external operator such that `dpdu.arguments() = (v,)`. 

But now if we consider dp’(cos(u))[v]  = dp * -sin(u) * v, we end up with `dp[(cos(u),); -sin(u)*v]` and dp.arguments() = (-sin(u)*v,) , so we need to hit the product handler.

What is a bit misleading here is that `arguments` refers to the expression on which we apply the action and not exclusively to the argument contained in this expression.

ufl/core/external_operator.py
nbouziani: We don’t need this, that should have been removed! I removed it.

ufl/core/external_operator.py
nbouziani: See comment above

ufl/core/external_operator.py
nbouziani: First, because when form compiling is not hit because of caching we don’t create the derivatives of `p`. So that is why we attach the information about what was made out of `p` to the kernel (cf. https://github.com/firedrakeproject/ufl/blob/af22ffd3583fef39621a879eb707f9a1b5e7e34a/ufl/algorithms/compute_form_data.py#L368)

Second, because we want to guarantee consistency across the coefficients associated with the derivatives of `p`. If we reconstruct a derivative of `p` that has already been made, we don’t need to make a new_object we just go through the dependency and returns the one with the correct derivatives multi-index (cf. https://github.com/firedrakeproject/ufl/blob/e9ca185c60e7d8344b5ba61000d8a88a9d13a742/ufl/core/external_operator.py#L198).

nbouziani:
> Thanks @dham , I like the concept as well as @meg-simula . I am just missing a pure UFL example code, @nbouziani provided `point_expr` which is a black box wrt. UFL. I will have a look at the code after announced updates.

The updates have been pushed

ufl/core/external_operator.py
wence-: OK, I think I understand. However, there are a few things that need to be tidied up.

1. The convention is that names starting with `_` are implementation detail (private). But here they are really providing public API. So can you make that change and document things.

2. Can you break the reference cycles (probably using `weakref.ref` or `weakref.proxy`)? I think the "parent" external operator needs to hold strong references to the child operators, but the children only need to hold weakrefs to the parent.

3. What does this in-place modification do to things like stability of the signature? Is it actually _necessary_, or is it a caching-based "performance" optimisation? For example, does all of your implementation require that differentiating the same extop twice with the same derivative multiindices give you back something that is object-identical, or just compares equal?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+        :param operands: operands on which acts the :class:`ExternalOperator`.
+        :param function_space: the :class:`.FunctionSpace`,
+            or :class:`.MixedFunctionSpace` on which to build this :class:`Function`.
+            Alternatively, another :class:`Coefficient` may be passed here and its function space
+            will be used.
+        :param derivatives: tuple scecifiying the derivative multiindex.
+        :param coefficient: ufl.Coefficient associated to the ExternalOperator representing what is produced by the operator
+        :param arguments: tuple composed of tuples whose first argument is a ufl.Argument or ufl.Expr
+            containing several ufl.Argument objects and whose second arguments is a boolean indicating
+            whether we take the action of the adjoint: used when we take the action.
```
wence-, line 44: This describes what the argument is, but is rather opaque. When do you take the action? Why like this with a bunch of boolean flags?

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+@ufl_type(num_ops="varying", inherit_indices_from_operand=0, is_differential=True)
+class ExternalOperator(Operator):
+
+    # Slots are disabled here because they cause trouble in PyDOLFIN
+    # multiple inheritance pattern:
+    _ufl_noslots_ = True
+
+    _external_operator_type = ''
```
wence-, line 31: What is this thing? It seems like it is used to specify certain behaviour of the operator. But it's just a random string that sometimes has a magic meaning.

Two options: 

1. make `ExternalOperator` a base class and the types inherit, e.g. `GlobalExternalOperator` etc..., then the handlers can deal with them

2. Use an enum and make sure the handlers explicitly handle all the cases 

```python
from enum import IntEnum

OperatorTypes(IntEnum):
   GLOBAL = 0
   LOCAL = 1

class ExternalOperator(Operator):
    Type = OperatorTypes

    operator_type = None

...

class ...(MultiFunction):
    def external_operator(...):
        typ = o.operator_type
        if not isinstance(typ, Enum) or typ not in o.Type:
           raise ValueError("Unknown operator type")
        if typ == o.Type.GLOBAL:
           ...
```

I think I prefer the former.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+        else:
+            original_function_space = ref_coefficient.ufl_function_space()
+            self.derivatives = (0,) * len(self.ufl_operands)
+
+        if coefficient is None:
+            coefficient = Coefficient(original_function_space)
+        elif not isinstance(coefficient, Coefficient):
+            raise TypeError('Expecting a Coefficient and not %s', type(coefficient))
+        self._coefficient = coefficient
+        self._original_function_space = self.coefficient.ufl_function_space()
```
wence-, line 93: You should probably check that the coefficient has a function space that matches the one you computed.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+
+        self._extop_dependencies = [self, ]
+
+        if self.derivatives == (0,) * len(self.ufl_operands):
+            self._extop_master = self
+
+    @property
+    def coefficient(self):
+        return self._coefficient
```
wence-, line 102: Normal UFL convention is that these are zero-argument functions, not properties.

wence-: And indeed, you inconsistently apply that convention (compare `self.arguments()` but `self.coefficient`).

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+            # TODO: subclass MixedElement to be able to do that !
+            raise NotImplementedError("MixedFunctionSpaces not handled yet")
+
+        if len(sub_element.sub_elements()) != 0:
+            sub_element = sub_element.sub_elements()[0]
+
+        if len(s) == 0:
+            ufl_element = sub_element
+        elif len(s) == 1:
+            ufl_element = VectorElement(sub_element)
```
wence-, line 178: Ignores `s`, when it should probably say `dim=s[0]`.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+        if deriv_multiindex != self.derivatives:
+            # If we are constructing a derivative
+            self._extop_master._extop_dependencies.append(reconstruct_op)
+            reconstruct_op._extop_master = self._extop_master
+        else:
+            reconstruct_op._extop_master = self._extop_master
+            reconstruct_op._extop_dependencies = [e._ufl_expr_reconstruct_(*operands) for e in self._extop_dependencies[1:]]
+        return reconstruct_op
+
+    def _add_dependencies(self, derivatives, args):
```
wence-, line 219: I think this is part of the public API, and therefore needs a docstring.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+        if self.derivatives == (0,) * len(self.ufl_operands):
+            self._extop_master = self
+
+    @property
+    def coefficient(self):
+        return self._coefficient
+
+    def _extract_coeffs_and_args(self, operands):
+        from ufl.algorithms.analysis import extract_arguments
+        args = tuple((e, is_adjoint) for e, is_adjoint in operands if len(extract_arguments(e)) != 0)
```
wence-, line 106: Given the docstring of what the `arguments` argument to `ExternalOperator` is, I don't understand how `extract_arguments(e)` can ever return a len-0 thing.

ufl/core/external_operator.py
```diff
@@ -0,0 +1,254 @@
...
+        ops = tuple((e, is_adjoint) for e, is_adjoint in operands if (e, is_adjoint) not in args)
+        return ops, args
+
+    def arguments(self):
+        return self._arguments
+
+    def action_args(self):
+        return self._action_args
```
wence-, line 114: What are these things?

wence- (changes requested):
More comments for clarification/fixing.

jhale:
First of all, I am generally positive about this PR, I see the need for this functionality (a generic mapping/operator/function from Expressions to Coefficients, and support for AD).

One thing I'd question is the naming `ExternalOperator`, particularly the `External` prefix. Could you explain the reasoning behind `External` a bit further? I understand that concretely in Firedrake you eventually want to generate a call to External code (outside UFL). But we have use cases where we might want to `ufl.substitute` certain terms for manually derived UFL Expressions. Possible alternatives for naming might be: `AbstractOperator`, `NotionalOperator`, `Map`. It would be good to keep things in UFL as generic and 'mathematical' as possible.

dham:
Hi Jack,

It's "external" in the sense that it's external to UFL. It's a foreign function interface. I think that that's a better description than the other suggestions you make, which I think mean different things. I think I don't understand your use case - if you want to express your operator in UFL then you can already do that, right?

jedbrown:
Is it intended that this capability could be used to implement an exact or Roe Riemann solver for DG? (These involve a Newton solve for a "star" state and an eigendecomposition, respectively.) What if the equation of state is implicit `f(pressure; density, energy) = 0`?

dham (commented):
> Is it intended that this capability could be used to implement an exact or Roe Riemann solver for DG? (These involve a Newton solve for a "star" state and an eigendecomposition, respectively.) What if the equation of state is implicit `f(pressure; density, energy) = 0`?

Hi Jed. At UFL level either of these are supported, because UFL is purely symbolic. This PR sets up a symbolic operator whose evaluation is defined outside of UFL. The operator behaves symbolically correctly, for example when differentiated with respect to one of its arguments.

Actual calculation requires a matching implementation in the problem solving environment (e.g. Firedrake or FEniCS). That implementation needs be able to evaluate the operator and appropriate derivatives and/or derivative actions. We have a draft implementation in Firedrake for implicit functions like the one you describe. We haven't done an operator with the right stencil for a Roe solver, but it would be possible and wouldn't require different code at the UFL level.

jedbrown:
Thanks, could you point me at the Firedrake example you mentioned?

mscroggs:
Can we close this now that #181 is open?

jhale:
@nbouziani So where is the post dual PR?

PR #27, opened 2020-05-27
Group integrals after degree estimation

michalhabera:
This PR first runs degree estimation on integrals, which attaches estimated polynomial degree as a integrand's metadata. It groups integrals only after.

This has several performance implications. E.g. in the current master
`u*v*dx + inner(grad(u), grad(v))*dx` would be grouped into one integral `(u*v + inner(grad(u), grad(v)))*dx` for which one polynomial degree is estimated. Note this degree is too high for the mass operator term.

With this PR it becomes for `P_k` functions (and affine geometry)
`u*v*dx(metadata={"est_degree": k**2}) + inner(grad(v), grad(v))*dx(metadata={"est_degree": (k-1)**2})`
and these integrals **won't** be later grouped by UFL.

### DIFF
--- a/ufl/algorithms/compute_form_data.py
+++ b/ufl/algorithms/compute_form_data.py
@@ -264,18 +264,18 @@ def compute_form_data(form,
     # user-defined coefficient relations it just gets too messy
     form = apply_derivatives(form)
 
+    # Estimate polynomial degree of integrands now, before applying
+    # any pullbacks and geometric lowering.  Otherwise quad degrees
+    # blow up horrifically.
+    if do_estimate_degrees:
+        form = attach_estimated_degrees(form)
+
     # --- Group form integrals
     # TODO: Refactor this, it's rather opaque what this does
     # TODO: Is self.original_form.ufl_domains() right here?
     #       It will matter when we start including 'num_domains' in ufc form.
     form = group_form_integrals(form, self.original_form.ufl_domains(),
                                 do_append_everywhere_integrals=do_append_everywhere_integrals)
-
-    # Estimate polynomial degree of integrands now, before applying
-    # any pullbacks and geometric lowering.  Otherwise quad degrees
-    # blow up horrifically.
-    if do_estimate_degrees:
-        form = attach_estimated_degrees(form)
 
     if do_apply_function_pullbacks:
         # Rewrite coefficients and arguments in terms of their


dham:
Is this actually a Good Thing? I see that you end up over integrating the derivative term, but the benefit is a single quadrature loop. If you think, in particular, about the non-affine element case where Jacobian has to be calculated at every quadrature point, it's not obvious to me that this would be a benefit.

michalhabera:
I agree that the question when would this actually produce a "faster" (whatever that means) code is extremely case dependent.
But this change is not restrictive. If you write your integral as `(whatever1 + whatever2)*dx` then this will get you the same quadrature degree (UFL won't split the integrand).
If you decide to split it, then both terms get their optimal degrees.

While with the current master, if you'd like to tell UFL to not merge integrals with the same integral data you'd need to put there some dummy metadata, to prevent UFL from grouping it, e.g.
```python
u*v*dx(metadata={"hello": "there"}) + inner(grad(u), grad(v))*dx(metadata={"hello": "here"})
```

michalhabera:
One more thought to add is that this PR just preserves more information. You can easily, later, in a form compiler decide if several integrals which differ only in quadrature scheme should be grouped based on some more compiler-specific logic. And you can group them easily.

It is more problematic to try to split integrands after there were grouped. I don't recall any mechanism in UFL for this, other than traversing the tree and looking for `Sum`?

dham:
Well yes, but that's not what current form compilers do, and it's a subtle change that could easily affect performance. I think that at a minimum there needs  to be an extended transition to allow form compiler authors to work out the correct response. This could be achieved by adding a switch onto the `compute_form_data` API. We already do that for lots of things.

michalhabera:
On a side note, I do not like the bloated interface of `compute_form_data`. It suggests to me a wrong design. There are too many things happening and controlling them with handful of kwargs sounds wrong. Compilers should rather do each individual operation if they decide to, based on their heuristic.

As a temporary fix, I can add `do_group_integral=True`, this would allow us in FFCx to do grouping manually if needed...

dham:
I agree that compute_form_data is a mess. However we are where we are and I think that the extra argument is the best short term fix.

michalhabera:
Agreed with @dham to choose a less invasive approach for now.

PR #30, opened 2020-06-08
Support point/vertex cells

ReubenHill:
Makes changes needed to support elements with vertex cells.

Note that the changes made to `determinant_expr` in `ufl/compound_expressions.py` to make the determinant of `Zero` be `zero()` were, ultimately, not used because of the change made to `ufl/algorithms/apply_integral_scaling.py`. I think they are technically correct but am happy to remove them if they seem incongruous.

### DIFF
--- a/ufl/finiteelement/mixedelement.py
+++ b/ufl/finiteelement/mixedelement.py
@@ -56,9 +56,12 @@ class MixedElement(FiniteElementBase):
 
         # Check that all elements use the same quadrature scheme TODO:
         # We can allow the scheme not to be defined.
-        quad_scheme = elements[0].quadrature_scheme()
-        if not all(e.quadrature_scheme() == quad_scheme for e in elements):
-            error("Quadrature scheme mismatch for sub elements of mixed element.")
+        if len(elements) == 0:
+            quad_scheme = None
+        else:
+            quad_scheme = elements[0].quadrature_scheme()
+            if not all(e.quadrature_scheme() == quad_scheme for e in elements):
+                error("Quadrature scheme mismatch for sub elements of mixed element.")
 
         # Compute value sizes in global and reference configurations
         value_size_sum = sum(product(s.value_shape()) for s in self._sub_elements)
@@ -292,10 +295,8 @@ class VectorElement(MixedElement):
         # Initialize element data
         MixedElement.__init__(self, sub_elements, value_shape=value_shape,
                               reference_value_shape=reference_value_shape)
-        # FIXME: Storing this here is strange, isn't that handled by
-        # subclass?
-        self._family = sub_element.family()
-        self._degree = sub_element.degree()
+        FiniteElementBase.__init__(self, sub_element.family(), cell, sub_element.degree(), quad_scheme,
+                                   value_shape, reference_value_shape)
         self._sub_element = sub_element
 
         # Cache repr string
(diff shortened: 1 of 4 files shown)


ufl/finiteelement/mixedelement.py
```diff
@@ -292,8 +295,8 @@ def __init__(self, family, cell=None, degree=None, dim=None,
         # Initialize element data
         MixedElement.__init__(self, sub_elements, value_shape=value_shape,
                               reference_value_shape=reference_value_shape)
-        # FIXME: Storing this here is strange, isn't that handled by
-        # subclass?
+        FiniteElementBase.__init__(self, sub_element.family(), cell, degree, quad_scheme,
+                                   value_shape, reference_value_shape)
         self._family = sub_element.family()
         self._degree = sub_element.degree()
         self._sub_element = sub_element
```
dham, line 302: Check this.

ufl/finiteelement/mixedelement.py
```diff
@@ -56,9 +56,12 @@ def __init__(self, *elements, **kwargs):
...
         # We can allow the scheme not to be defined.
-        quad_scheme = elements[0].quadrature_scheme()
-        if not all(e.quadrature_scheme() == quad_scheme for e in elements):
-            error("Quadrature scheme mismatch for sub elements of mixed element.")
+        if len(elements) == 0:
+            quad_scheme = None
+        else:
+            quad_scheme = elements[0].quadrature_scheme()
+            if not all(e.quadrature_scheme() == quad_scheme for e in elements):
+                error("Quadrature scheme mismatch for sub elements of mixed element.")
```
wence-, line 64: Having to do this suggests that one should revisit the design decision of making `VectorElement` and `TensorElement` subclasses of `MixedElement`. Although I don't suggest we do so here, can you open an issue proposing that we split them out? I would advocate keeping `MixedElement` only for a "bag of unrelated elements" and create new `Vector/TensorElement` classes that rank-augment an existing `FiniteElement` object. One could then go one step further and just make `VectorElement` a functional constructor for a `TensorElement`.

ufl/finiteelement/mixedelement.py
ReubenHill: I've changed this such that `FiniteElementBase.__init__` sets everything except `self._sub_element`. (see rebased commit 93a9194e9a2fdac6b5c4101f4c4005a9fcbaa7db ) Note that it now sets `self.degree` to `sub_element.degree()` in `FiniteElementBase.__init__`.

ufl/finiteelement/mixedelement.py
```diff
@@ -56,9 +56,12 @@ def __init__(self, *elements, **kwargs):
 
         # Check that all elements use the same quadrature scheme TODO:
         # We can allow the scheme not to be defined.
-        quad_scheme = elements[0].quadrature_scheme()
-        if not all(e.quadrature_scheme() == quad_scheme for e in elements):
-            error("Quadrature scheme mismatch for sub elements of mixed element.")
+        if len(elements) == 0:
```
michalhabera, line 59: Can you explain why this condition is needed for point cell?

ReubenHill: Vertex cells have topological dimension zero, so, thanks to this snippet
https://github.com/FEniCS/ufl/blob/3492e727f4ae3e5f2bf2d6a51e0ac37903fa9978/ufl/finiteelement/mixedelement.py#L286
the list of `sub_elements` which enters the `MixedElement` constructor ends ups being an empty list

ufl/finiteelement/mixedelement.py
ReubenHill: See #33

ReubenHill:
@michalhabera happy to merge this? I don't have the necessary permissions.

PR #32, opened 2020-06-26
Direct serendipity

rckirby:
Some elements may be defined directly on each physical cell.  This PR 

- introduces a "physical" element mapping
- updates some differentiation rules for that special case
- adds a new element "Sdirect" that is intended to be bound to Todd Arbogast's "directly-defined" serendipity family for general non-affine quadrilaterals.  

(Note: An implementation of this element lives in branches of tsfc and finat in Firedrake with PR to be opened soon depending on this one)


ufl/finiteelement/elementlist.py
```diff
@@ -259,6 +259,9 @@ def show_elements():
 register_element("Gauss-Lobatto-Legendre Edge L2", "GLL-Edge L2", 0, L2, "L2 Piola", (0, None),
                  ("interval",))
 
+# directly-defined serendipity elements ala Arbogast
+register_element2("Sdirect", 0, H1, "physical", (1, None), ("quadrilateral",))
```
michalhabera, line 263: Should this be rather defined on `cubes`, not just quad?

ufl/algorithms/apply_function_pullbacks.py
```diff
@@ -64,10 +64,14 @@ def apply_single_function_pullbacks(g):
     mapping = element.mapping()
 
     r = ReferenceValue(g)
-
     gsh = g.ufl_shape
     rsh = r.ufl_shape
 
+    if element.mapping() == "physical":
```
michalhabera, line 70: `if mapping == "physical":`

ufl/algorithms/apply_function_pullbacks.py
```diff
@@ -64,10 +64,14 @@ def apply_single_function_pullbacks(g):
...
+        # TODO: Is this right for immersed things
```
michalhabera, line 71: This should be okay for k-surfaces in n-spaces, k<n, as far as the basis for your physically mapped elements is "scalar valued". AFAIK this is only an issue for tensor valued basis like Hdiv or Hdivdiv...

ufl/finiteelement/elementlist.py
rckirby: Todd tells me there is an extension to hexes, but they have to have straight faces (which is not true unless the four vertices of a face are coplanar), and I don't know if it is written up.  I have only implemented quads.

ufl/algorithms/apply_function_pullbacks.py
rckirby: fixed.

ufl/algorithms/apply_function_pullbacks.py
rckirby: It is unclear what the right thing is here.  Serendipity spaces are very fussy under mapping, and I have no idea if they can be defined on or mapped to an immersed manifold.

ufl/finiteelement/elementlist.py
rckirby: We plan working on some quadrilateral H(div) families, but the implementation is not finished yet.  Should we register them now so we don't need a separate future request to ufl?

rckirby:
Are we happy with this?

ufl/algorithms/apply_function_pullbacks.py
wence-: I will update the comments to clarify this point.

ufl/finiteelement/elementlist.py
```diff
@@ -259,6 +259,13 @@ def show_elements():
 register_element("Gauss-Lobatto-Legendre Edge L2", "GLL-Edge L2", 0, L2, "L2 Piola", (0, None),
                  ("interval",))
 
+# directly-defined serendipity elements ala Arbogast
+# currently the theory is only really worked out for quads.
+register_element2("Sdirect", 0, H1, "physical", (1, None), ("quadrilateral",))
+
+register_element2("Sdirect_Hdiv_full", 1, HDiv, "physical", (1, None), ("quadrilateral",))
+register_element2("Sdirect_Hdiv_reduced", 1, HDiv, "physical", (1, None), ("quadrilateral",))
```
miklos1, line 267: These names look odd. Normally, one uses spaces in long UFL element names, not underscores.  I think they should look more like titles than variable names.

miklos1: c.f. `"Nedelec 1st kind H(div)"`, `"Extended-Gauss-Legendre Edge L2"`, or `"Q- Lambda L2"`

PR #34, opened 2020-07-07
Add check_restrictions in compute_form_data

cdaversin:
Update existing check_restrictions code and use it in compute_form_data to limit usage of restrictions to interior facet and custom integral types.

### DIFF
--- a/ufl/algorithms/check_restrictions.py
+++ b/ufl/algorithms/check_restrictions.py
@@ -11,6 +11,8 @@ from ufl.log import error
 from ufl.corealg.multifunction import MultiFunction
 from ufl.corealg.map_dag import map_expr_dag
 
+from ufl.classes import Restricted
+
 
 class RestrictionChecker(MultiFunction):
     def __init__(self, require_restriction):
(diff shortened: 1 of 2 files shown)


ufl/algorithms/check_restrictions.py
```diff
@@ -11,15 +11,34 @@
...
-        pass
+        # Check is given expression involves restrictions
+        self.find_restriction(o)
+        # Restrictions are needed and only allowed for interior facet integrals
+        if self.require_restriction:
+            if self.current_restriction is None:
+                error("Form argument must be restricted in interior facet integrals.")
+        else:
+            if self.current_restriction is not None:
+                error("Restrictions are only allowed for interior facet and custom integral types.")
```
wence-, line 41: This hand-coded recursion in `find_restriction` unfortunately isn't a scalable approach [it's quadratic in tree depth]. Unfortunately, to fix this is a little bit of work, the `map_dag` infrastructure provides facility for DAG-based visiting and memoization of post-order traversal (child before parent), but not the preorder (parent before child) traversal we need here. I think this is a wider problem.

I think you can phrase this visit as post-order traversal, but it's a little fiddly.

Unrelated to this patch, reading this code I wondered how it works at all, since in `def restricted` it calls `self.visit` but `visit` is not defined anywhere.

I think this was not completely ported from the old `ReuseTransformer` infrastructure.

It's not previously been used in the normal pipeline (we have some error checking via `apply_default_restrictions` that is only applied to interior_facet integrals), so this is why we haven't ever noticed.

PR #38, opened 2020-08-18
Some zany H(div) elements

rckirby:
We are adding support for three nonstandard H(div) finite elements -- the vector-valued Mardal-Tai-Winther triangle (also an H^1 nonconforming element) and two tensor-valued elements due to Arnold and Winther (conforming and nonconforming versions).  Separate PR is coming for FIAT.


test/test_sobolevspace.py
```diff
@@ -138,13 +137,15 @@ def test_contains_hdiv():
...
         FiniteElement("BDFM", triangle, 2),
+        
         # HDiv elements:
         HDiv(TensorProductElement(FiniteElement("DG", triangle, 1),
                                   FiniteElement("CG", interval, 2))),
         HDiv(TensorProductElement(FiniteElement("RT", triangle, 1),
                                   FiniteElement("DG", interval, 1))),
         HDiv(TensorProductElement(FiniteElement("N1curl", triangle, 1),
                                   FiniteElement("DG", interval, 1)))
+
```
miklos1, line 148: unnecessary whitespace changes in this hunk

ufl/finiteelement/elementlist.py
```diff
@@ -145,6 +144,10 @@ def show_elements():
 register_element("HDiv Trace", "HDivT", 0, L2, "identity", (0, None), any_cell)
 register_element("Hellan-Herrmann-Johnson", "HHJ", 2, HDivDiv,
                  "double contravariant Piola", (0, None), ("triangle",))
+register_element("Nonconforming Arnold-Winther", "AWnc", 2, HDivDiv,
+                 "double contravariant Piola", (0, None), ("triangle", "tetrahedron"))
+register_element("Conforming Arnold-Winther", "AWc", 2, HDivDiv,
+                 "double contravariant Piola", (0, None), ("triangle", "tetrahedron"))
```
miklos1, line 150: FIAT code says that degree must be 3 for conforming, and 2 for nonconforming AW elements.  Here it is 0 to infinity.  That is inconsistent.

rckirby: The conforming element is defined for orders >= 3 (depending on how you count...).  The nonconforming element is only defined based on (constrained) quadratics.  

The FIAT implementation for the conforming case is currently only degree 3, but this could be generalized at some point if desired.  So I will change the nonconforming case to (2, 2) and the conforming for (3, None) so we won't have to change UFL if FIAT gets an upgrade here.

miklos1 (approved):
This is okay now.

PR #40, opened 2020-09-03
Allow Q1 calculation of CellDiameter to work for tensor product cells

pefarrell:
Without this, `CellDiameter` doesn't work on meshes constructed via extrusion.


pefarrell:
I should also say that this patch was proposed by @wence- .

wence- (approved):

michalhabera:
Wouldn't a line extruded into quad extruded into hex have degree `(1, (1,1))`?

miklos1:
> Wouldn't a line extruded into quad extruded into hex have degree `(1, (1,1))`?

I don't think Firedrake supports double extrusion.  If you want a hex, you need to start with an unstructured quad.

PR #42, opened 2020-09-21
use complex-coordinate-space distance in GeometryLoweringApplier

ksagiyam:
This PR is to use complex-coordinate-space distance in `GeometryLoweringApplier` to make this class safe to use in complex mode.

### DIFF
--- a/ufl/algorithms/apply_geometry_lowering.py
+++ b/ufl/algorithms/apply_geometry_lowering.py
@@ -37,7 +37,7 @@ from ufl.classes import (Expr, Form, Integral,
 # FacetOrientation, QuadratureWeight,
 
 from ufl.tensors import as_tensor, as_vector
-from ufl.operators import sqrt, max_value, min_value
+from ufl.operators import sqrt, max_value, min_value, conj, real
 
 from ufl.compound_expressions import determinant_expr, cross_expr, inverse_expr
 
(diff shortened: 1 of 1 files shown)


michalhabera:
Why is the `abs()` needed? Note `a*conj(a) >= 0`.

dham:
> Why is the `abs()` needed? Note `a*conj(a) >= 0`.

I think the abs is needed to force the type of the expression to real. I guess you could also use `real` to achieve the same effect. Note the `a*conj(a)` still has type complex, and might have a small imaginary part due to roundoff.

ufl/algorithms/apply_geometry_lowering.py
```diff
@@ -238,7 +238,7 @@ def circumradius(self, o):
         edges = CellEdgeVectors(domain)
         num_edges = edges.ufl_shape[0]
         j = Index()
-        elen = [sqrt(edges[e, j] * edges[e, j]) for e in range(num_edges)]
+        elen = [sqrt(abs(edges[e, j] * conj(edges[e, j]))) for e in range(num_edges)]
```
wence-, line 241: I think that we should advertise that all of these things come out as `real`. Since they are sometimes later used in a context where that is required (e.g. computing the minimum of the edge lengths below).

So I would prefer that all of these are written as: 

```python
[real(sqrt(real(edges[e, j] * conj(edges[e, j])))) for e in range(num_edges)]
```

etc...

ksagiyam:
Thanks, everyone. Wrapped `sqrt` with `abs`.

wence- (approved):
Thanks, looks OK (and is a no-op in real mode).

PR #43, opened 2020-09-22
Fix registration of MTW elements again

wence-:
Value-rank and mapping were both reported incorrectly in a botched
merge. Fix and add a test to check them.


wence-:
Now I am happy that this is working, sorry for the churn.

PR #46, opened 2021-01-25
Add degree argument to tensor product element reconstruction

rckirby:
In using low-order preconditioners, we need to reconstruct a finite element with a lower-degree instance of the same element family.  This PR enables this for tensor product elements.

### DIFF
--- a/ufl/finiteelement/tensorproductelement.py
+++ b/ufl/finiteelement/tensorproductelement.py
@@ -17,7 +17,8 @@ from ufl.log import error
 from ufl.cell import TensorProductCell, as_cell
 from ufl.sobolevspace import DirectionalSobolevSpace
 
-from ufl.finiteelement.finiteelementbase import FiniteElementBase, FiniteElement
+from ufl.finiteelement.finiteelementbase import FiniteElementBase
+from ufl.finiteelement.finiteelement import FiniteElement
 
 from collections.abc import Iterable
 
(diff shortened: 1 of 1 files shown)


ufl/finiteelement/tensorproductelement.py
```diff
@@ -103,8 +106,19 @@ def sub_elements(self):
...
         return self._sub_elements
 
-    def reconstruct(self, cell=None):
-        return TensorProductElement(*self.sub_elements(), cell=cell)
+    def reconstruct(self, cell=None, degree=None):
+        if degree is None:
+            return TensorProductElement(*self.sub_elements(), cell=cell)
+        elif isinstance(degree, int):
+            return TensorProductElement(*[FiniteElement(subel.family(),
+                                                        subel.cell(),
```
wence-, line 114: Ignores the provided cell argument

ufl/finiteelement/tensorproductelement.py
```diff
@@ -103,8 +106,19 @@ def sub_elements(self):
...
+                                                        degree)
+                                          for subel in self.sub_elements()])
+        elif isinstance(degree, Iterable):
+            return TensorProductElement(*[FiniteElement(subel.family(),
+                                                        subel.cell(),
```
wence-, line 119: Likewise here

ufl/finiteelement/tensorproductelement.py
```diff
@@ -103,8 +106,19 @@ def sub_elements(self):
...
+                                                        deg)
+                                          for (subel, deg) in zip(self.sub_elements(), degree)])
```
wence-, line 122: Should probably check that the lengths match

wence- (changes requested):

rckirby:
Pablo's [PMGSNES PR](https://github.com/firedrakeproject/firedrake/pull/1963) over on Firedrake does a lot of what I need here, and it makes sense to work in one place at a time.  His is more general, and we should try to migrate that over here once it lands.  So I'm converting this to draft for the time being.

PR #47, opened 2021-02-04
Register useful testing element

ReubenHill:
Register a new element with integral moment nodes and point evaluation nodes on an interval cell which is very useful for sanity checking. Name is very much up for negotiation! 

See also https://github.com/FEniCS/fiat/pull/64

### DIFF
--- a/ufl/finiteelement/elementlist.py
+++ b/ufl/finiteelement/elementlist.py
@@ -99,6 +99,7 @@ any_cell = (None,
 
 # Elements in the periodic table # TODO: Register these as aliases of
 # periodic table element description instead of the other way around
+register_element("NewElement", "NewElement", 0, H1, "identity", (3, 3), simplices[0])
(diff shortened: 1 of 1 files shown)


ufl/finiteelement/elementlist.py
```diff
@@ -150,6 +150,7 @@ def show_elements():
                  "double contravariant Piola", (2, 2), ("triangle", "tetrahedron"))
 register_element("Conforming Arnold-Winther", "AWc", 2, HDivDiv,
                  "double contravariant Piola", (3, None), ("triangle", "tetrahedron"))
+register_element("C0 Modified", "C0Mod", 0, H1, "identity", (3, 3), simplices[0])
```
wence-, line 153: Prefer not to do this at all, the "variant" mechanism can allow us to specify the difference.

ReubenHill:
I've decided to move this into the test itself and not create a public element (see https://github.com/firedrakeproject/firedrake/pull/1971) so will close this.

PR #49, opened 2021-02-23
WithMapping element to override pullback selection

colinjcotter:
Provide capability to override pullbacks e.g. Piola maps,
needed when the element is extruded.

### DIFF
--- a/ufl/finiteelement/hdivcurl.py
+++ b/ufl/finiteelement/hdivcurl.py
@@ -86,3 +86,31 @@ class HCurlElement(FiniteElementBase):
     def shortstr(self):
         "Format as string for pretty printing."
         return "HCurlElement(%s)" % str(self._element.shortstr())
+
+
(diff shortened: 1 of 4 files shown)


ufl/finiteelement/hdivcurl.py
```diff
@@ -86,3 +86,31 @@ def __str__(self):
     def shortstr(self):
         "Format as string for pretty printing."
         return "HCurlElement(%s)" % str(self._element.shortstr())
+
+
+class WithMapping(FiniteElementBase):
+    def __init__(self, wrapee, mapping):
```
wence-, line 92: Can you add a docstring here describing what this does?

colinjcotter: done

michalhabera:
Why not just `original_element._mapping = "identity"` etc.?

wence- (approved):

PR #55, opened 2021-05-19
Add functions to strip and attach data references in forms

connorjward:
This change adds two new functions: `strip_terminal_data` and `replace_terminal_data`.
The former takes a form as input and returns a new 'stripped' form, where the terminals have been replaced with UFL-only equivalents, and a mapping to allow one to reconstruct the original form with `replace_terminal_data`.

These functions will make it easier to cache forms in Firedrake and other libraries that use UFL.
Currently caching forms is an issue because they contain subclassed UFL objects where data has been attached to them (e.g. a Firedrake `Function`).
The advantage to producing 'stripped' forms is that they do not hold these data references and can be safely cached without introducing a memory leak.

### DIFF
--- a/ufl/algorithms/__init__.py
+++ b/ufl/algorithms/__init__.py
@@ -37,6 +37,8 @@ __all__ = [
     "expand_derivatives",
     "extract_coefficients",
     "strip_variables",
+    "strip_terminal_data",
+    "replace_terminal_data",
     "post_traversal",
     "change_to_reference_grad",
     "expand_compounds",
@@ -93,6 +95,8 @@ from ufl.algorithms.transformer import Transformer, ReuseTransformer
 # from ufl.algorithms.transformer import is_post_handler
 from ufl.algorithms.transformer import apply_transformer
 from ufl.algorithms.transformer import strip_variables
+from ufl.algorithms.strip_terminal_data import strip_terminal_data
+from ufl.algorithms.strip_terminal_data import replace_terminal_data
 # from ufl.algorithms.replace import Replacer
 from ufl.algorithms.replace import replace
 from ufl.algorithms.change_to_reference import change_to_reference_grad
--- a/ufl/domain.py
+++ b/ufl/domain.py
@@ -209,6 +209,9 @@ class TensorProductMesh(AbstractDomain):
     def ufl_cell(self):
         return self._ufl_cell
 
+    def ufl_meshes(self):
+        return self._ufl_meshes
+
     def is_piecewise_linear_simplex_domain(self):
         return False  # TODO: Any cases this is True
 
--- a/test/test_strip_forms.py
+++ b/test/test_strip_forms.py
@@ -1,0 +1,106 @@
+import gc
+import sys
+
+import pytest
+
+from ufl import *
+from ufl.algorithms import strip_terminal_data, replace_terminal_data
+from ufl.core.ufl_id import attach_ufl_id
+from ufl.core.ufl_type import attach_operators_from_hash_data
+
+
+MIN_REF_COUNT = 2
+"""The minimum value returned by sys.getrefcount."""
+
+
+@attach_operators_from_hash_data
+@attach_ufl_id
+class AugmentedMesh(Mesh):
+    def __init__(self, *args, data):
+        super().__init__(*args)
+        self.data = data
+
+
+@attach_operators_from_hash_data
+class AugmentedFunctionSpace(FunctionSpace):
+    def __init__(self, *args, data):
+        super().__init__(*args)
+        self.data = data
+
+
+class AugmentedCoefficient(Coefficient):
+    def __init__(self, *args, data):
+        super().__init__(*args)
+        self.data = data
+
+
+class AugmentedConstant(Constant):
+    def __init__(self, *args, data):
+        super().__init__(*args)
+        self.data = data
+
+
+def test_strip_form_arguments_strips_data_refs():
+    mesh_data = object()
+    fs_data = object()
+    coeff_data = object()
+    const_data = object()
+
+    # Sanity check
+    assert sys.getrefcount(mesh_data) == MIN_REF_COUNT
+    assert sys.getrefcount(fs_data) == MIN_REF_COUNT
+    assert sys.getrefcount(coeff_data) == MIN_REF_COUNT
+    assert sys.getrefcount(const_data) == MIN_REF_COUNT
+
+    cell = triangle
+    domain = AugmentedMesh(cell, data=mesh_data)
+    element = FiniteElement("Lagrange", cell, 1)
+    V = AugmentedFunctionSpace(domain, element, data=fs_data)
+
+    v = TestFunction(V)
+    u = TrialFunction(V)
+    f = AugmentedCoefficient(V, data=coeff_data)
+    k = AugmentedConstant(V, data=const_data)
+
+    form = k*f*inner(grad(v), grad(u))*dx
+
+    # Remove extraneous references
+    del cell, domain, element, V, v, u, f, k
+
+    assert sys.getrefcount(mesh_data) == MIN_REF_COUNT + 1
+    assert sys.getrefcount(fs_data) == MIN_REF_COUNT + 1
+    assert sys.getrefcount(coeff_data) == MIN_REF_COUNT + 1
+    assert sys.getrefcount(const_data) == MIN_REF_COUNT + 1
+
+    stripped_form, mapping = strip_terminal_data(form)
+
+    del form, mapping
+    gc.collect()  # This is needed to update the refcounts
+
+    assert sys.getrefcount(mesh_data) == MIN_REF_COUNT
+    assert sys.getrefcount(fs_data) == MIN_REF_COUNT
+    assert sys.getrefcount(coeff_data) == MIN_REF_COUNT
+    assert sys.getrefcount(const_data) == MIN_REF_COUNT
+
+
+def test_strip_form_arguments_does_not_change_form():
+    mesh_data = object()
+    fs_data = object()
+    coeff_data = object()
+    const_data = object()
+
+    cell = triangle
+    domain = AugmentedMesh(cell, data=mesh_data)
+    element = FiniteElement("Lagrange", cell, 1)
+    V = AugmentedFunctionSpace(domain, element, data=fs_data)
+
+    v = TestFunction(V)
+    u = TrialFunction(V)
+    f = AugmentedCoefficient(V, data=coeff_data)
+    k = AugmentedConstant(V, data=const_data)
+
+    form = k*f*inner(grad(v), grad(u))*dx
+    stripped_form, mapping = strip_terminal_data(form)
+
+    assert stripped_form.signature() == form.signature()
+    assert replace_terminal_data(stripped_form, mapping) == form
--- a/ufl/algorithms/strip_terminal_data.py
+++ b/ufl/algorithms/strip_terminal_data.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+"""Algorithm for replacing form arguments with 'stripped' versions where any
+data-carrying objects have been extracted to a mapping."""
+
+from ufl.classes import Form, Integral
+from ufl.classes import Argument, Coefficient, Constant
+from ufl.classes import FunctionSpace, TensorProductFunctionSpace, MixedFunctionSpace
+from ufl.classes import Mesh, MeshView, TensorProductMesh
+from ufl.algorithms.replace import replace
+from ufl.corealg.map_dag import map_expr_dag
+from ufl.corealg.multifunction import MultiFunction
+
+
+class TerminalStripper(MultiFunction):
+    def __init__(self):
+        super().__init__()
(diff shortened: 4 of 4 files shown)


ufl/core/terminal.py
```diff
@@ -102,3 +102,9 @@ class FormArgument(Terminal):
 
     def __init__(self):
         Terminal.__init__(self)
+
+    def ufl_strip_data(self):
+        """Return a copy of the form argument with all non-UFL information
+        removed. This is useful for removing references to large data structures
+        so that the object is cachable."""
+        raise NotImplementedError("ufl_strip_data should be implemented by the subclass")
```
wence-, line 110: I think I would prefer if these removal of data just lived as a handler on the transformer.

ufl/algorithms/transformer.py
```diff
@@ -216,11 +216,35 @@ def variable(self, o):
...
+class FormArgumentReplacer(CopyTransformer):
+    def __init__(self, arg_map):
+        CopyTransformer.__init__(self)
+        self.arg_map = arg_map
+
+    def terminal(self, o):
+        try:
+            return self.arg_map[o]
+        except KeyError:
+            raise ValueError(f"FormArgument {o} is not present in the mapping")
```
wence-, line 239: You weren't to know this, but the right thing to inherit from is not this `Transformer` but rather the `MultiFunction` in `ufl.corealg.multifunction`.

This one can almost certainly just be implemented with `ufl.replace`.

The stripper could too I think as:

```
ufl.replace(form, dict(zip(o, o.ufl_strip_data()) for o in form.terminals()))
```

(Pseudo-code).

ufl/algorithms/map_integrands.py
```diff
@@ -19,20 +19,24 @@
...
         nonzero_integrals = [itg for itg in mapped_integrals
                              if not isinstance(itg.integrand(), Zero)]
         return Form(nonzero_integrals)
     elif isinstance(form, Integral):
         itg = form
         if (only_integral_type is None) or (itg.integral_type() in only_integral_type):
-            return itg.reconstruct(function(itg.integrand()))
+            integrand = function(itg.integrand())
+            domain = integrand.ufl_domain() if reconstruct_integral_domain else None
+            return itg.reconstruct(integrand, domain=domain)
```
wence-, line 39: You need this because the domain may also have "heavy" data?

ufl/domain.py
```diff
@@ -104,6 +108,9 @@ def ufl_coordinate_element(self):
     def ufl_cell(self):
         return self._ufl_coordinate_element.cell()
 
+    def ufl_strip_data(self):
+        return Mesh(self._ufl_coordinate_element, self._ufl_id, self._ufl_cargo)
```
wence-, line 112: I think canonically the cargo is heavy.

ufl/domain.py
```diff
@@ -154,6 +161,9 @@ def ufl_cell(self):
     def is_piecewise_linear_simplex_domain(self):
         return self._ufl_mesh.is_piecewise_linear_simplex_domain()
 
+    def ufl_strip_data(self):
+        return MeshView(self._ufl_mesh, self._topological_dimension, self._ufl_id)
```
wence-, line 165: If you're doing this, surely you need to strip the mesh.

ufl/domain.py
```diff
@@ -212,6 +222,10 @@ def ufl_cell(self):
     def is_piecewise_linear_simplex_domain(self):
         return False  # TODO: Any cases this is True
 
+    def ufl_strip_data(self):
+        meshes = (mesh.ufl_strip_data() for mesh in self._ufl_meshes)
+        return TensorProductMesh(meshes, self._ufl_id)
```
wence-, line 227: Is it bad that the ids are preserved for the meshes?

ufl/algorithms/__init__.py
```diff
@@ -37,6 +37,8 @@
     "expand_derivatives",
     "extract_coefficients",
     "strip_variables",
+    "strip_form_arguments",
```
michalhabera, line 40: Shouldn't it be named `strip_form_data` etc.?

ufl/algorithms/map_integrands.py
connorjward: Yep.

ufl/algorithms/__init__.py
connorjward: I called it this because ' form data' has a different meaning in UFL already (`ufl/algorithms/formdata.py`) so I didn't want a conflict. Perhaps I could call it `strip_form_argument_data`?

ufl/domain.py
connorjward: Good spot. Thanks.

ufl/domain.py
connorjward: I don't know. This was a somewhat arbitrary choice I made to make the two objects as similar as possible.

ufl/algorithms/map_integrands.py
connorjward: This isn't a very appealing fix though because I now have this `reconstruct_integral_domain` keyword argument hanging around. I initially tried without but then `test/test_mixed_function_space.py::test_mixed_functionspace` was failing with `ufl.log.UFLException: Found domains with different geometric dimensions`.

ufl/domain.py
connorjward: Do we know what type `cargo` is? I don't know whether I should just pass it as `None` here or to implement a `ufl_strip_data` for it.

ufl/core/terminal.py
connorjward: Yes that would definitely be neater.

ufl/domain.py
connorjward: The docstring for it says:
> `"Return carried object that will not be used by UFL."`

so I will leave it out.

ufl/core/terminal.py
connorjward: This is now done.

ufl/domain.py
```diff
@@ -209,6 +209,9 @@ def ufl_coordinate_element(self):
     def ufl_cell(self):
         return self._ufl_cell
 
+    def ufl_meshes(self):
```
connorjward, line 212: This is a missing part of the API. I only added it so I wouldn't have to access a private attribute when stripping one of these objects. What I have done mirrors how private attributes are accessed throughout UFL.

ufl/algorithms/map_integrands.py
wence-: This seems like a bug though, because AIUI your change doesn't touch anything that should affect that test?

ufl/algorithms/strip_form_argument_data.py
```diff
@@ -0,0 +1,97 @@
...
+    the objects to be replaced and the values are the objects they are
+    to be replaced with.
+
+    Note that this function is simply the opposite of strip_form_arguments.
+    """
+    if not isinstance(form, Form):
+        raise ValueError(f"{form} must be a Form")
+
+    return map_integrand_dags(Replacer(mapping), form,
+                              reconstruct_integral_domain=True)
```
wence-, line 68: What is the difference between this function and `replace`?

ufl/algorithms/strip_form_argument_data.py
```diff
@@ -0,0 +1,97 @@
...
+    "Return a new function space with all non-UFL information removed."
+    if isinstance(function_space, FunctionSpace):
+        return FunctionSpace(_strip_domain(function_space.ufl_domain()),
+                             function_space.ufl_element())
+    elif isinstance(function_space, TensorProductFunctionSpace):
+        return TensorProductFunctionSpace(_strip_function_space(sub)
+                                          for sub in function_space.ufl_sub_spaces())
+    elif isinstance(function_space, MixedFunctionSpace):
+        return MixedFunctionSpace(_strip_function_space(sub)
+                                  for sub in function_space.ufl_sub_spaces())
```
wence-, line 81: Need to splat this generator I think.

ufl/algorithms/strip_form_argument_data.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def _strip_function_space(function_space):
+    "Return a new function space with all non-UFL information removed."
+    if isinstance(function_space, FunctionSpace):
+        return FunctionSpace(_strip_domain(function_space.ufl_domain()),
+                             function_space.ufl_element())
+    elif isinstance(function_space, TensorProductFunctionSpace):
+        return TensorProductFunctionSpace(_strip_function_space(sub)
+                                          for sub in function_space.ufl_sub_spaces())
```
wence-, line 78: Need to splat this generator too.

ufl/algorithms/map_integrands.py
```diff
@@ -19,20 +19,24 @@
...
                             for itg in form.integrals()]
         nonzero_integrals = [itg for itg in mapped_integrals
                              if not isinstance(itg.integrand(), Zero)]
         return Form(nonzero_integrals)
     elif isinstance(form, Integral):
         itg = form
         if (only_integral_type is None) or (itg.integral_type() in only_integral_type):
-            return itg.reconstruct(function(itg.integrand()))
+            integrand = function(itg.integrand())
+            domain = integrand.ufl_domain() if reconstruct_integral_domain else None
```
wence-, line 38: Arguably this reconstruction of the domain is always the right thing to do?

wence- (commented):
Some queries/nitpicks

ufl/algorithms/__init__.py
wence-: FWIW, I don't like "arguments" since that suggests that coefficients are not stripped.

How about `detach_terminals` ?

wence-:
Can you rebase and squash appropriately?

ufl/algorithms/strip_form_argument_data.py
connorjward: The `reconstruct_integral_domain` argument. I didn't want to add a keyword argument to `replace`.

ufl/algorithms/map_integrands.py
connorjward: That's possible. I'll investigate.

ufl/algorithms/__init__.py
connorjward: I don't think 'detach' is right because that doesn't make it clear that the terminals are getting replaced. `strip_terminal_data`?

ufl/algorithms/strip_form_argument_data.py
connorjward: But if the `test_mixed_function_space` error does turn out to be a bug then I can abandon `reconstruct_integral_domain` and this function can go.

ufl/algorithms/map_integrands.py
connorjward: I agree. It just resulted in the test failure. If I fix that then that's what I'll do.

ufl/algorithms/map_integrands.py
connorjward: So the issue here is that in `test_mixed_functionspace.py` we have forms defined over multiple domains (1D, 2D and 3D). This means that just calling `integrand.ufl_domain()` fails since there is more than one domain being considered. I have no real idea what to do about this. @wence- any thoughts?

connorjward:
@wence- I've finally gotten around to getting this done. Could you take a look?

garth-wells:
The names of the function seem to imply mutation of UFL objects, and I can't quiet decipher from the PR description whether or not this is the case?

dham:
> The names of the functions seem to imply mutation of UFL objects, and I can't quite decipher from the PR description whether or not this is the case?

The docstrings make it explicit that new UFL objects are returned. Can you suggest names you would prefer?

garth-wells:
> > The names of the functions seem to imply mutation of UFL objects, and I can't quite decipher from the PR description whether or not this is the case?
> 
> The docstrings make it explicit that new UFL objects are returned. Can you suggest names you would prefer?

It would help if the PR description and docstrings were clearer. The docstring for `strip_domain` is clear, but it isn't in the PR description. `strip_form_argument_data` is in the PR description but I don't see it in the code changes. Like wise `attach_form_argument_data`.

`strip_terminal_data` and `replace_terminal_data` docstrings would be clearer if they started "Create a new UFL expression from ......" rather than "Convert an UFL expression to a ....".

connorjward:
> > > The names of the functions seem to imply mutation of UFL objects, and I can't quite decipher from the PR description whether or not this is the case?
> > 
> > 
> > The docstrings make it explicit that new UFL objects are returned. Can you suggest names you would prefer?
> 
> It would help if the PR description and docstrings were clearer. The docstring for `strip_domain` is clear, but it isn't in the PR description. `strip_form_argument_data` is in the PR description but I don't see it in the code changes. Like wise `attach_form_argument_data`.
> 
> `strip_terminal_data` and `replace_terminal_data` docstrings would be clearer if they started "Create a new UFL expression from ......" rather than "Convert an UFL expression to a ....".

Sorry I forgot to update the PR description, I've done that now. I have also amended the docstrings so they should make it clearer that no mutation is going on.

garth-wells (approved):

PR #59, opened 2021-05-23
Add pyramid

jmv2009:
Add pyramid

(following https://github.com/FEniCS/ufl/commit/4cc3c59fce84000eafd29297cad5abb99175379a)


chrisrichardson:
Should also be fine, though some things will not work, e.g. facet integrals... `cellname2facetname` needs revising, for example...

Issue #62, opened 2021-07-03
`preprocessed_form` contains global derivatives of the Jacobian


jpdean:
The following code
```
import ufl

coord_element = ufl.VectorElement("Lagrange", ufl.quadrilateral, 1)
mesh = ufl.Mesh(coord_element)
element = ufl.FiniteElement("RTCF", ufl.quadrilateral, 1)
V = ufl.FunctionSpace(mesh, element)
v = ufl.TrialFunction(V)
form = ufl.inner(1, ufl.div(v)) * ufl.dx

form_data = ufl.algorithms.compute_form_data(
    form,
    do_apply_function_pullbacks=True,
    do_apply_integral_scaling=True,
    do_apply_geometry_lowering=True,
    preserve_geometry_types=(ufl.classes.Jacobian,),
    do_apply_restrictions=True,
    do_append_everywhere_integrals=False)

print(form_data.preprocessed_form)
```
produces an output containing `grad(J)` instead of transforming the Jacobian derivatives to `reference_grad(J)`. Is this the desired behaviour? Should these global derivatives not have been transformed to reference derivatives after preprocessing? I discovered this whilst investigating https://github.com/FEniCS/ffcx/issues/373.

PR #63, opened 2021-07-05
Transform global derivatives of Jacobian to reference derivatives in `preprocessed_form`

jpdean:
This PR fixes https://github.com/FEniCS/ufl/issues/62 and https://github.com/FEniCS/ffcx/issues/373.

### DIFF
--- a/ufl/algorithms/apply_derivatives.py
+++ b/ufl/algorithms/apply_derivatives.py
@@ -501,7 +501,7 @@ class GradRuleset(GenericDerivativeRuleset):
             # form compilers will handle this.
             domain = o.ufl_domain()
             K = JacobianInverse(domain)
-            Do = self._grad_to_reference_grad(o, K)
+            Do = grad_to_reference_grad(o, K)
             return Do
 
     def jacobian_inverse(self, o):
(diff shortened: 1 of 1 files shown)


ufl/algorithms/apply_derivatives.py
```diff
@@ -593,6 +591,25 @@ def _grad(self, o):
         # 1) n = count number of Grads, get f
         # 2) if not f.has_derivatives(n): return zero(...)
 
+    def _grad_to_reference_grad(self, o, K):
```
wence-, line 594: Nitpick, this function doesn't need `self` I think, so could it just be a free function?

ufl/algorithms/apply_derivatives.py
```diff
@@ -491,24 +491,26 @@ def __init__(self, geometric_dimension):
...
-        """Default for geometric quantities is dg/dx = 0 if piecewise constant, otherwise keep Grad(g).
+        """Default for geometric quantities is do/dx = 0 if piecewise constant,
+        otherwise transform derivatives to reference derivatives.
         Override for specific types if other behaviour is needed."""
         if is_cellwise_constant(o):
             return self.independent_terminal(o)
         else:
             # TODO: Which types does this involve? I don't think the
             # form compilers will handle this.
-            return Grad(o)
```
wence-, line 501: I think you can also remove the `TODO` comment?

wence- (commented):
Thanks! Modulo minor comments I think this looks right.

ufl/algorithms/apply_derivatives.py
jpdean: Sure! I put it there because it is only used by the `GradRuleset` class, but I'm happy to move it

garth-wells (approved):

PR #66, opened 2021-08-13
Catch users who try to index UFL expressions with negative indices

pefarrell:
I had code where I wanted to access the last component of a vector expression, and used `expr[-1]`. It turns out that UFL indices don't wrap-around, and it just caused an obscure error in the form compiler later. This patch catches such user errors at a more useful stage.


wence- (approved):

wence-:
Thanks

PR #67, opened 2021-08-25
Fix the docstring for dot

lrtfm:
Fix the docstring for dot


dham (approved):
Obviously correct. Thanks for fixing this.

PR #68, opened 2021-08-27
Refactor pullback application

wence-:
Application of pullback to expressions had duplicate code, and didn't handle TensorElements with symmetries unless the sub element was identity-mapped. Refactor and generalise to handle both of those cases.

### DIFF
--- a/test/test_apply_function_pullbacks.py
+++ b/test/test_apply_function_pullbacks.py
@@ -1,32 +1,34 @@
 #!/usr/bin/env py.test
 # -*- coding: utf-8 -*-
 
-from pytest import raises
+import numpy
 from ufl import *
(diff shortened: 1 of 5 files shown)


wence-:
OK, now ready. As a heads up, this will probably somewhat change the preprocessed form that ffc sees. I think in many cases it should produce marginally better code (since it doesn't lower to unrolled loops immediately).

IgorBaratta:
> OK, now ready. As a heads up, this will probably somewhat change the preprocessed form that ffc sees. I think in many cases it should produce marginally better code (since it doesn't lower to unrolled loops immediately).

I will check what the impacts on ffcx are.
I was trying to revert some unrolled loops (pullbacks) in ffcx.
But yes,  it's better to do that on the UFL side.

IgorBaratta (approved):

PR #74, opened 2021-12-01
Parse multiple expressions

michalhabera:
This expects `expressions` gobal name in UFL files be list of expressions (ufl.core.expr.Expr and evaluation points). Needed to parse multiple expressions on FFCx side. Not affecting FEniCS(old) nor Firedrake.


ufl/algorithms/formfiles.py
```diff
@@ -204,7 +204,7 @@ def get_form(name):
     # Validate types
     if not isinstance(ufd.expressions, (list, tuple)):
         error("Expecting 'expressions' to be a list or tuple, not '%s'." % type(ufd.expressions))
-    if not all(isinstance(e, Expr) for e in ufd.expressions):
+    if not all(isinstance(e[0], Expr) for e in ufd.expressions):
```
wence-, line 207: I guess I don't know how the form files work, but should one not write:

```
all(isinstance(e, Expr) for e in itertools.chain(*ufd.expressions))
```

Rather than implicitly indicating that the `ufd.expressions` variable is a list of singleton lists.

michalhabera: Currently, "expressions" is a list of tuples of different type, `List[Tuple(ufl.core.expr.Expression, numpy.ndarray)]`, but this is not standardised within UFL. We only need to know that first item in the tuple is of UFL Expression type. The second item in the tuple is used to tell what reference points expression is evaluated.

wence- (approved):

PR #76, opened 2021-12-23
Add function to get the `ufl.Cell` corresponding to a cell's facet

jpdean:
In some applications, it is useful to get the `ufl.Cell` corresponding to a cell's facets (i.e. for a triangle in 2D space, the facets are `ufl.Cell('interval', 2)`).

### DIFF
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -16,6 +16,7 @@
 import numbers
 import functools
 
+import ufl.cell
 from ufl.log import error
 from ufl.core.ufl_type import attach_operators_from_hash_data
 
@@ -169,6 +170,11 @@ class Cell(AbstractCell):
         # This is used in geometry.py
         fn = cellname2facetname[self.cellname()]
         return num_cell_entities[fn][1]
+
+    def facet_cell(self):
+        "The ufl.Cell representing the facets of self."
+        return ufl.Cell(cellname2facetname[self.cellname()],
+                        self.geometric_dimension())
 
     # --- Special functions for proper object behaviour ---
 
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -166,27 +166,17 @@ class Cell(AbstractCell):
 
     # --- Facet properties ---
 
-    def facet_cells(self):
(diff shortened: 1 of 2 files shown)


chrisrichardson:
I've a feeling this already exidts as a free function, something like `cell2facet` - it's problematic with e.g. prism cells, because it depends which facet... can we do something better?

chrisrichardson:
I am thinking of `ufl.cell.cellname2facetname()` which is used elsewhere in ffcx. It would be good to clean this up, rather than have two similar functions.

chrisrichardson:
Looking at your PR, I see now that was a silly comment of mine, as you are using this function already. I'm still not happy about making use of something that is pretty flawed.

jpdean:
OK, thanks! I'll look into improving `cellname2facetname`.

chrisrichardson:
This looks good to me. @dham - can we delete `ufl.cell.cellname2facetname`? Or is this used in firedrake?

chrisrichardson (approved):
This is an improvement, which will allow us to remove the oddly named and inconsistent `cellname2facetname` function.

ufl/cell.py
```diff
@@ -164,11 +166,27 @@ def num_facets(self):
 
     # --- Facet properties ---
 
-    def num_facet_edges(self):
-        "The number of facet edges."
-        # This is used in geometry.py
-        fn = cellname2facetname[self.cellname()]
-        return num_cell_entities[fn][1]
+    def facet_cells(self):
```
garth-wells, line 169: Would `facet_types` be a better name?

ufl/cell.py
```diff
@@ -164,11 +166,27 @@ def num_facets(self):
...
+        "A list of ufl.Cell representing the facets of self."
+        # TODO Replace if with match-case from Python 3.10
+        cell_name = self.cellname()
+        if cell_name == "interval":
+            facet_names = ["vertex"]
```
garth-wells, line 174: Returning a tuple would probably be better than a list.

PR #77, opened 2022-01-10
Remove cellname2facetname

jpdean:
Once https://github.com/FEniCS/ffcx/pull/435 has been merged, `cellname2facetname` can be removed from UFL.

### DIFF
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -93,6 +93,7 @@ num_cell_entities = {"vertex": (1,),
 # Mapping from cell name to topological dimension
 cellname2dim = dict((k, len(v) - 1) for k, v in num_cell_entities.items())
 
+
 # --- Basic cell representation classes
 
 @attach_operators_from_hash_data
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
Just a heads up that this will break old FFC, in particular `ffc/representationutils.py`. A simple fix is to copy the definition of `cellname2facetname` there. (maybe ping @cdaversin if she is working on FFC and @drew-parsons if he ends up with this error while preparing ubuntu packages)

jhale:
We are no longer maintaining FFC, if people want to submit patches they are of course welcome to and we can merge them.

The Ubuntu packages should be built from the last stable release 2019.1.0 + build patches? So there should be no issue.

drew-parsons:
It's a small enough change, looks like it shouldn't be too hard to patch into ffc.

jorgensd:
> We are no longer maintaining FFC, if people want to submit patches they are of course welcome to and we can merge them.
> 
> The Ubuntu packages should be built from the last stable release 2019.1.0 + build patches? So there should be no issue.

I guess the main problem is that we should really have a final release of DOLFIN, which includes the MeshView functionality that @cdaversin added after the release of 2019.1.0.

PR #79, opened 2022-01-14
Modernise Python packaging

jhale:
This enables modern isolated builds via `pyproject.toml`, shifts `setup.py` information into `setup.cfg`, keeps a barebones `setup.py` for those who use editable installs, and creates an action for creating sdists and wheels of UFL and uploading them to pypi. Dependencies for linting, building docs, tests and running CI are moved to central `extras_require` section, e.g. `python -m pip install .[test]`.


garth-wells (approved):

setup.cfg
```diff
@@ -1,3 +1,59 @@
+[metadata]
+name = fenics-ufl
+version = 2019.2.0.dev0
```
francesco-ballarin, line 3: 2019 should probably be 2021, as there is a 2021.1.0 tag out

PR #80, opened 2022-01-14
Modernize parts of codebase and remove deprecated code-paths

jorgensd:
This PR is a starting point for modernizing the ufl code base.

We start by introducing:
- [Type hints PEP 484](https://www.python.org/dev/peps/pep-0484/)
- [fstrings PEP 498](https://www.python.org/dev/peps/pep-0498/) 

Also several functions that has been marked as deprecated has been removed:
- `Argument.ufl_domain()`
- `Argument.ufl_domains()`
- `Argument.ufl_element()`

### DIFF
--- a/ufl/algebra.py
+++ b/ufl/algebra.py
@@ -177,7 +177,7 @@ class Product(Operator):
 
     def __str__(self):
         a, b = self.ufl_operands
-        return " * ".join((parstr(a, self), parstr(b, self)))
+        return f"{parstr(a, self)} * {parstr(b, self)}"
 
 
 @ufl_type(num_ops=2,
@@ -241,8 +241,7 @@ class Division(Operator):
         return e
 
     def __str__(self):
-        return " / ".join([parstr(self.ufl_operands[0], self),
-                           parstr(self.ufl_operands[1], self)])
+        return f"{parstr(self.ufl_operands[0], self)} / {parstr(self.ufl_operands[1], self)}"
 
 
 @ufl_type(num_ops=2,
@@ -299,7 +298,7 @@ class Power(Operator):
 
     def __str__(self):
         a, b = self.ufl_operands
-        return " ** ".join([parstr(a, self), parstr(b, self)])
+        return f"{parstr(a, self)} ** {parstr(b, self)}"
 
 
 @ufl_type(num_ops=1,
@@ -422,4 +421,4 @@ class Imag(Operator):
 
     def __str__(self):
         a, = self.ufl_operands
-        return "Im[%s]" % (parstr(a, self),)
+        return f"Im[{parstr(a, self)}]"
--- a/ufl/algebra.py
+++ b/ufl/algebra.py
@@ -1,6 +1,3 @@
-# -*- coding: utf-8 -*-
-"Basic algebra operations."
-
 # Copyright (C) 2008-2016 Martin Sandve Alnæs
 #
 # This file is part of UFL (https://www.fenicsproject.org)
@@ -8,6 +5,8 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 #
 # Modified by Anders Logg, 2008
+
+"Basic algebra operations."
 
 from ufl.log import error
 from ufl.core.ufl_type import ufl_type
@@ -89,24 +88,6 @@ class Sum(Operator):
 
     def __str__(self):
         ops = [parstr(o, self) for o in self.ufl_operands]
-        if False:
-            # Implementation with line splitting:
-            limit = 70
-            delimop = " + \\\n    + "
-            op = " + "
-            s = ops[0]
-            n = len(s)
-            for o in ops[1:]:
-                m = len(o)
-                if n + m > limit:
-                    s += delimop
-                    n = m
-                else:
-                    s += op
-                    n += m
-                s += o
-            return s
-        # Implementation with no line splitting:
         return "%s" % " + ".join(ops)
 
 
@@ -125,7 +106,7 @@ class Product(Operator):
         # Make sure everything is scalar
         if a.ufl_shape or b.ufl_shape:
             error("Product can only represent products of scalars, "
-                  "got\n\t%s\nand\n\t%s" % (ufl_err_str(a), ufl_err_str(b)))
+                  f"got\n\t {ufl_err_str(a)}\nand\n\t {ufl_err_str(b)}")
 
         # Simplification
         if isinstance(a, Zero) or isinstance(b, Zero):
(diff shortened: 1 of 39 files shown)


ufl/argument.py
```diff
@@ -38,97 +40,74 @@ class Argument(FormArgument):
...
         FormArgument.__init__(self)
 
-        if isinstance(function_space, FiniteElementBase):
-            # For legacy support for .ufl files using cells, we map the cell to
-            # the default Mesh
-            element = function_space
-            domain = default_domain(element.cell())
-            function_space = FunctionSpace(domain, element)
-        elif not isinstance(function_space, AbstractFunctionSpace):
+        if not isinstance(function_space, AbstractFunctionSpace):
```
jorgensd, line 47:
```suggestion
        if isinstance(function_space, FiniteElementBase):
            # For legacy support for .ufl files using cells, we map the cell to
            # the default Mesh
            element = function_space
            domain = default_domain(element.cell())
            function_space = FunctionSpace(domain, element)
        elif not isinstance(function_space, AbstractFunctionSpace):
```

ufl/algorithms/analysis.py
```diff
@@ -143,8 +143,10 @@ def extract_arguments_and_coefficients(a):
 
 def extract_elements(form):
     "Build sorted tuple of all elements used in form."
-    args = chain(*extract_arguments_and_coefficients(form))
-    return tuple(f.ufl_element() for f in args)
+    arguments, coefficients = extract_arguments_and_coefficients(form)
+    elements_a = tuple(a.ufl_function_space().ufl_element() for a in arguments)
+    elements_c = tuple(c.ufl_element() for c in coefficients)
+    return elements_a + elements_c
```
wence-, line 149: Why don’t coefficients have a function space?

ufl/algorithms/apply_derivatives.py
```diff
@@ -545,13 +545,21 @@ def argument(self, o):
     def reference_value(self, o):
         # grad(o) == grad(rv(f)) -> K_ji*rgrad(rv(f))_rj
         f = o.ufl_operands[0]
-        if f.ufl_element().mapping() == "physical":
+        if hasattr(f, "ufl_function_space"):
+            el = f.ufl_function_space().ufl_element()
+        else:
+            el = f.ufl_element()
+        if el.mapping() == "physical":
```
wence-, line 552: Why does this backwards compat (?) need to be introduced?

ufl/argument.py
```diff
@@ -54,81 +59,63 @@ def __init__(self, function_space, number, part=None):
...
-    def ufl_domain(self):
-        "Deprecated, please use .ufl_function_space().ufl_domain() instead."
-        # TODO: deprecate("Argument.ufl_domain() is deprecated, please
-        # use .ufl_function_space().ufl_domain() instead.")
-        return self._ufl_function_space.ufl_domain()
-
-    def ufl_element(self):
-        "Deprecated, please use .ufl_function_space().ufl_element() instead."
-        # TODO: deprecate("Argument.ufl_domain() is deprecated, please
-        # use .ufl_function_space().ufl_element() instead.")
```
wence-, line 84: TBH, I think this deprecation comment seems misguided. I think that there should be a consistent way to get the domain from any expression, and it should probably be `ufl_domains`. Otherwise you have to special case arguments everywhere.

ufl/domain.py
```diff
@@ -339,7 +339,11 @@ def extract_domains(expr):
     "Return all domains expression is defined on."
     domainlist = []
     for t in traverse_unique_terminals(expr):
-        domainlist.extend(t.ufl_domains())
+        if hasattr(t, "ufl_function_space"):
+            domains = t.ufl_function_space().ufl_domains()
+        else:
+            domains = t.ufl_domains()
```
wence-, line 345: See, this seems like a backwards step.

wence- (changes requested):
I think most of the cleanup looks fine, but the removal of ufl_domain from arguments looks like a step in the wrong direction.

ufl/algorithms/analysis.py
jorgensd: It does. A mistake on my behalf.

jorgensd:
> I think most of the cleanup looks fine, but the removal of ufl_domain from arguments looks like a step in the wrong direction.

I've reverted the removal of `ufl_domains` (but still removed `ufl_element` and `ufl_domain`. 
I've also cleaned up the code using the fact that coefficients also have a `ufl_function_space`.

ufl/algebra.py
```diff
@@ -318,7 +299,7 @@ def evaluate(self, x, mapping, component, index_values):
 
     def __str__(self):
         a, b = self.ufl_operands
-        return "%s ** %s" % (parstr(a, self), parstr(b, self))
```
jhale, line 321: Result looks different to old code?

jorgensd: It should give the same result. MWE:
```python
import ufl

cell = ufl.triangle
el = ufl.VectorElement("Lagrange", cell, 1)
mesh = ufl.Mesh(el)
a = ufl.Constant(mesh)
b = ufl.Constant(mesh)
c = a ** b
print(c)
```
yielding 
```bash
c_0 ** c_1
```

jhale: You're right! I haven't used join a great deal. Is it really clearer than an f-string or format? I can see the advantage for many arguments, but 2 seems marginal.

jorgensd: > You're right! I haven't used join a great deal. Is it really clearer than an f-string or format? I can see the advantage for many arguments, but 2 seems marginal.

I agree it is marginal. I’ll just revert it.

ufl/algorithms/apply_derivatives.py
```diff
@@ -545,13 +545,16 @@ def argument(self, o):
...
             # TODO: Do we need to be more careful for immersed things?
             return ReferenceGrad(o)
 
         if not f._ufl_is_terminal_:
             error("ReferenceValue can only wrap a terminal")
-        domain = f.ufl_domain()
+
+        domains = f.ufl_function_space().ufl_domains()
+        domain = None if len(domains) == 0 else domains[0]
```
wence-, line 557: I think it should be the case that a function space always has a domain? Should we instead write this as:

```
domain, = f.ufl_function_space().ufl_domains()
```

To assert that there is only one.

ufl/algorithms/analysis.py
```diff
@@ -143,8 +143,9 @@ def extract_arguments_and_coefficients(a):
 
 def extract_elements(form):
     "Build sorted tuple of all elements used in form."
+    arguments, coefficients = extract_arguments_and_coefficients(form)
```
wence-, line 146: this seems unneeded now?

ufl/algorithms/compute_form_data.py
```diff
@@ -60,8 +60,9 @@ def _compute_element_mapping(form):
     #   worked around to drop this requirement
 
     # Extract all elements and include subelements of mixed elements
-    elements = [obj.ufl_element() for obj in chain(form.arguments(),
-                                                   form.coefficients())]
+    elements_a = tuple(a.ufl_function_space().ufl_element() for a in form.arguments())
+    elements_c = tuple(c.ufl_element() for c in form.coefficients())
```
wence-, line 64: coefficients and arguments both have function spaces attached, so can both these iterations be merged?

ufl/argument.py
wence-: what was the resolution to this?

wence- (commented):
I'm a bit confused by some of the changes in this PR. Most of it is just some minor formatting changes for consistency, but then there are some actual API changes that I think warrant discussion in terms of whether they are a step forward (eg the stuff around arguments and domains). Can you split it in two to make it easier to review?

mscroggs:
We should maybe close this, as I've started doing this is a series of smaller PRs (starting with #131)

PR #86, opened 2022-01-18
Add workflow to run DOLFINx and FFCx tests

mscroggs:
When a PR is opened, the FFCx and DOLFINx tests will be run using the updated UFL.

This will make is easier to catch when changes to UFL might require changes to FFCx and DOLFINx.

I think it would be sensible to add a similar CI for Firedrake at some point, if someone from Firedrake wants to set something similar up.

### DIFF
--- a/.github/workflows/fenicsx-tests.yml
+++ b/.github/workflows/fenicsx-tests.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install test dependencies
         run: |
           sudo apt-get install -y graphviz libgraphviz-dev ninja-build pkg-config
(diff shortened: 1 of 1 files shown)


wence- (approved):
I think this looks good. The firedrake build is rather heavier than dolfin so we'll probably just pull a container rather than building from scratch.

It's not as urgent since we (for zenodo-release purposes) maintain a UFL fork that just points at known good UFL-main commits (this is currently updated manually, though with these changes we could update it automatically.

garth-wells (approved):

PR #87, opened 2022-02-01
Dualspace

dham:
This PR implements Dual function spaces, cofunctions in dual function spaces, arguments, and all the consequential objects needed to support dual operations and operators in UFL. This is documented in this thesis: https://drive.google.com/file/d/1hH_ntqZYThicpkStVwkBR-CrQ185bb95/view

It is intended to be backwards compatible so packages which don't currently use dual spaces will never see any dual objects.

This is an unfortunately big diff, but is claimed to be the minimal change which implements a closed set of functionality. The interpolation operator and external operators use this to provide actual functionality. They will follow in a subsequent PR.


michalhabera:
Thanks for the contribution. I had a quick look.

Why is it needed to distinguish primal and dual Coefficients with a new name "Cofunction"? I'd say
```python
V_dual = V.dual()
v_dual = Coefficient(V_dual)
```
(and the same for Argument) should have the same meaning. No need for that extra code which check if you ask for "cofunction" on primal or dual space. One can add the parent class runtime based on if FS is primal or dual. Also the word cofunction seems to be reserved a different meaning, https://en.wikipedia.org/wiki/Cofunction.

I like you've added `BaseForm` for a linear functional. The old `Form` should not have been reserved for "integral" forms, but that would require a larger braking change...

Can you explain why is an extra class `FormSum` required to handle the sum of `BaseForms`?

ReubenHill:
> Also the word cofunction seems to be reserved a different meaning, https://en.wikipedia.org/wiki/Cofunction.

Here it's being used to refer to the category theory meaning (hence the link at the top of the wikipedia article to https://en.wikipedia.org/wiki/Dual_(category_theory). But googling for "Cofunction" definitely gives you results that are all about angles!

dham:
Sorry for taking a while to get back, I'm catching up after being on strike.

The reason why dual coefficients and dual arguments need to be different types to primal ones is that they are fundamentally different objects with different algebraic properties. `Cofunctions` and `Coarguments` are `FormBase` objects they can be used where other 1-forms could be used, and they can't be used where `UFL.expr` is needed. For example, suppose `u` is an `Argument` and `v` is a `Cofunction`. The following makes mathematical sense and can be supported in code:
```python
u*dx + v
```
However, `v*dx` does not make sense and should be an error. Similarly, if `w` is a `Coefficient` then:
```python
u*dx + w
```
must be an error. The clean way to achieve this is through different types. What complicates things in this case is that the inheritance relationships work in two ways. `Coefficient` and `Cofunction` are both known values in vector spaces (primal and dual spaces respectively). Similarly, `Argument` and `Coargument` are both unknown values in vector spaces. The class inheritance relations reflect the commonality and differences between these cases cleanly. The alternative would be an enormous number of hand-coded dispatches on the type of the underlying function space. We should make the OO features of Python work for us and not fight against them.

On the name `Cofunction`, I am open to alternatives but I haven't heard a good one yet. Sure, we're overloading a term which is used in a different bit of maths, but the category-theoretic use of the prefix "co-" does make sense. The strictly UFL name would be `Cocoefficient`, however the confusion that that would cause is easily predictable.

The reason that we need `FormSum` is the same as the reason that we need `Action` and `Adjoint`. Every operation which is legal on a `FormBase` either has to evaluate eagerly or has to return a symbolic expression which can be evaluated at assembly time. Since taking the action, adjoint or a linear combination of `FormBase` objects can be a large numerical operation, it needs to be delayed to assembly time. It should be noted that `FormSum` is quite similar to `Form` in this respect: `Form` stores a sum of integrals. As you say, if we were doing this from scratch then we might implement and name `Form` differently. I think the design we have here is a good balance of keeping things as clean as we can without breaking what we already have.

ufl/form.py
```diff
@@ -550,3 +732,155 @@ def replace_integral_domains(form, common_domain):  # TODO: Move elsewhere
...
+        # warning("Calling repr on form is potentially expensive and should be avoided except during debugging.")
+        # Not caching this because it can be huge
+        itgs = ", ".join(repr(component) for component in self.components())
+        r = "FormSum([" + itgs + "])"
+        return r
+
+
+@ufl_type()
+class ZeroBaseForm(BaseForm):
+    """UFL base form type: respresents a zero base form
```
dham, line 841: typo

ufl/algorithms/ad.py
```diff
@@ -27,6 +28,12 @@ def expand_derivatives(form, **kwargs):
     if kwargs:
         warning("Deprecation: expand_derivatives no longer takes any keyword arguments")
 
+    if isinstance(form, Adjoint):
+        dform = expand_derivatives(form._form)
+        if dform == 0:
+            return dform
+        raise NotImplementedError('Adjoint derivative is not supported.')
```
dham, line 35: Put a comment in here reminding us that this is a 3-form can't happen situation.

ufl/finiteelement/elementlist.py
```diff
@@ -200,6 +200,10 @@ def show_elements():
                   ("quadrilateral",))
 register_element2("BDMCF", 1, HDiv, "contravariant Piola", (1, None),
                   ("quadrilateral",))
+register_element2("SminusE", 1, HCurl, "covariant Piola", (1, None), cubes[1:3])
+register_element2("SminusF", 1, HDiv, "contravariant Piola", (1, None), cubes[1:2])
+register_element2("SminusDiv", 1, HDiv, "contravariant Piola", (1, None), cubes[1:3])
+register_element2("SminusCurl", 1, HCurl, "covariant Piola", (1, None), cubes[1:3])
```
dham, line 206: rebase on FEniCS main.

nbouziani: Done

ufl/algorithms/ad.py
nbouziani: Done

mscroggs (approved):
I've made some minor comments/suggestions.

Once these are resolved, I'm happy for this to be merged.

setup.cfg
```diff
@@ -59,6 +59,8 @@ ci =
 
 [flake8]
 ignore = E501, W504,
+         W503,  # Line break before operator, no longer PEP8.
+         E129,  # Indentation, can trigger for valid code.
```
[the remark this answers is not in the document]
jorgensd: image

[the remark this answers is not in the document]
nbouziani: Thanks, I updated that and also the next one, which I think is also problematic:

>          E741                   # ambiguous variable name

ufl/core/ufl_type.py
```diff
@@ -372,3 +381,37 @@ def _ufl_expr_rbinop_(self, other):
         return cls
 
     return _ufl_type_decorator_
+
+
+class UFLType(type):
```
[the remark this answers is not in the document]
nbouziani: Thanks! Yes, things look a bit cleaner now.

ufl/finiteelement/elementlist.py
nbouziani: Yes - I removed them from that PR.

PR #94, opened 2022-03-19
Signature computation speedup

michalhabera:
EDIT: See comments for updates. Current solution is based on unique post traversal and computing hash at each node during the traversal.

For complicated UFL forms signature computation easily dominates JIT compilation, often even assembly times. Expression signatures are fetched e.g. in FFCx for disk caching - which contributes a unique hash into binary file name.

~This PR changes `pre_traversal` to `unique_pre_traversal` for expression hashdata, which has significant 700x speedup in some cases (see cProfiler images below). Additionally, operands (their ufl typecode) are added to the generated hashdata to make it unique.~

### DIFF
--- a/ufl/algorithms/signature.py
+++ b/ufl/algorithms/signature.py
@@ -122,7 +122,7 @@ def compute_expression_signature(expr, renumbering):  # FIXME: Fix callers
 
     # Pass it through a seriously overkill hashing algorithm
     # (should we use sha1 instead?)
-    return hashlib.sha512(expression_hashdata).hexdigest()
+    return expression_hashdata.hex()
 
 
 def compute_form_signature(form, renumbering):  # FIXME: Fix callers
--- a/ufl/algorithms/signature.py
+++ b/ufl/algorithms/signature.py
@@ -14,7 +14,7 @@ from ufl.classes import (Label,
                          GeometricQuantity, ConstantValue, Constant,
                          ExprList, ExprMapping)
 from ufl.log import error
-from ufl.corealg.traversal import traverse_unique_terminals, unique_pre_traversal
+from ufl.corealg.traversal import traverse_unique_terminals, unique_post_traversal
 from ufl.algorithms.domain_analysis import canonicalize_metadata
 
 
@@ -94,21 +94,18 @@ def compute_terminal_hashdata(expressions, renumbering):
 
 
 def compute_expression_hashdata(expression, terminal_hashdata):
-    expression_hashdata = []
+    cache = {}
 
-    for expr in unique_pre_traversal(expression):
-        # Visit each node only once, but store all its operands
+    for expr in unique_post_traversal(expression):
         if expr._ufl_is_terminal_:
-            expression_hashdata.append(terminal_hashdata[expr])
+            data = [terminal_hashdata[expr]]
         else:
-            expression_hashdata.append(expr._ufl_typecode_)
-            for op in expr.ufl_operands:
-                if op._ufl_is_terminal_:
-                    expression_hashdata.append(terminal_hashdata[op])
-                else:
-                    expression_hashdata.append(op._ufl_typecode_)
+            data = [expr._ufl_typecode_]
 
-    return expression_hashdata
+            for op in expr.ufl_operands:
+                data += cache[op]
+        cache[expr] = data
+    return cache[expression]
 
 
 def compute_expression_signature(expr, renumbering):  # FIXME: Fix callers
--- a/ufl/algorithms/signature.py
+++ b/ufl/algorithms/signature.py
@@ -14,7 +14,7 @@ from ufl.classes import (Label,
                          GeometricQuantity, ConstantValue, Constant,
                          ExprList, ExprMapping)
 from ufl.log import error
-from ufl.corealg.traversal import traverse_unique_terminals, pre_traversal
+from ufl.corealg.traversal import traverse_unique_terminals, unique_pre_traversal
 from ufl.algorithms.domain_analysis import canonicalize_metadata
 
 
@@ -98,7 +98,7 @@ def compute_expression_hashdata(expression, terminal_hashdata):
     # notation, i.e. we store the equivalent of '+ * a b * c d' for
     # the expression (a*b)+(c*d)
     expression_hashdata = []
-    for expr in pre_traversal(expression):
+    for expr in unique_pre_traversal(expression):
         if expr._ufl_is_terminal_:
             data = terminal_hashdata[expr]
         else:
--- a/ufl/algorithms/signature.py
+++ b/ufl/algorithms/signature.py
@@ -93,18 +93,21 @@ def compute_terminal_hashdata(expressions, renumbering):
     return terminal_hashdata
 
 
-def compute_expression_hashdata(expression, terminal_hashdata):
+def compute_expression_hashdata(expression, terminal_hashdata) -> bytes:
     cache = {}
 
     for expr in unique_post_traversal(expression):
+        # Uniquely traverse tree and hash each node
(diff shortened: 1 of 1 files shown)


ufl/algorithms/signature.py
```diff
@@ -94,19 +94,20 @@ def compute_terminal_hashdata(expressions, renumbering):
...
-            data = terminal_hashdata[expr]
+            expression_hashdata.append(terminal_hashdata[expr])
         else:
-            data = expr._ufl_typecode_  # TODO: Use expr._ufl_signature_data_()? More extensible, but more overhead.
-        expression_hashdata.append(data)
-    # Oneliner: TODO: Benchmark, maybe use a generator?
-    # expression_hashdata = [(terminal_hashdata[expr] if expr._ufl_is_terminal_ else expr._ufl_typecode_)
-    #                       for expr in pre_traversal(expression)]
+            expression_hashdata.append(expr._ufl_typecode_)
+            for op in expr.ufl_operands:
```
wence-, line 105: Why do you need to expand the first level of children for the signature? I would have thought that the unique pre-order visit is unique for signature purposes even without expanding the tree like this.

That is, why is this not a sufficient implementation?

```
return tuple(terminal_hashdata[expr] if expr._ufl_is_terminal_ else expr._ufl_typecode_ for expr in unique_pre_traversal(expression))
```
?

michalhabera: I did it to avoid the following
```python
import ufl
from ufl.corealg.traversal import unique_pre_traversal

c1 = ufl.Constant(ufl.triangle)
c2 = ufl.Constant(ufl.triangle)

expression = c2*c2 + c1
print(tuple(str(expr) if expr._ufl_is_terminal_ else expr._ufl_typecode_ for expr in unique_pre_traversal(expression)))

expression = c2*c1 + c1
print(tuple(str(expr) if expr._ufl_is_terminal_ else expr._ufl_typecode_ for expr in unique_pre_traversal(expression)))

```
which produces the same hashdata
```
(61, 62, 'c_1', 'c_0')
(61, 62, 'c_1', 'c_0')
```

However, when I think about it, there could be rare cases where even first level children expansion produces same expr hashdata. Any suggestions? We need to avoid non-unique traversal.

wence-: Could one do the following instead (I don't have the benchmarking setup to know if this is faster):

```python
class Hashdata(MultiFunction):
    def __init__(self, terminal_hashdata):
        super().__init__()
        self.terminal_hashdata = terminal_hashdata

    def expr(self, op, *ops):
        if op._ufl_is_terminal_:
            return (self.terminal_hashdata[op],)
        return tuple(op._ufl_typecode_, *ops)

def compute_expression_hashdata(expr, terminal_hashdata):
    mapper = Hashdata(terminal_hashdata)
    return map_expr_dag(mapper, expr)
```

michalhabera: It is worse, `MultiFunction` is applied with `map_expr_dag` which uses `unique_post_traversal`, but the overhead here I think comes from all the caching and dictionary access (and possibly hashing the elements too?) - see the profile

Screenshot 2022-03-30 at 14 59 18
.

wence-: Almost all of the cost looks to be hashing `FiniteElement` objects?

michalhabera: Probably due to `rcache` in `map_expr_dags` and `FiniteElementBase` hash not being cached.

I've managed to achieve ~4s with unique post traversal and insertion of operands based on cache (similar to map_expr_dags, but simpler). Almost all time spent is now in `str(that_huge_expressiondata_list)` in `signature.py` line 161.

The form I am using is a jacobian computed with UFL AD from a potential, where the potential is defined using closed-form expressions for eigenvalues of 3x3 Cauchy-Green tensor. We are essentially differentiating through roots of characteristic polynomial of ugly symbolic matrix.

wence-: I mean that's better, but still pretty bad. I wonder if it is faster to build the bytes array directly while traversing. `str` isn't taking advantage of the shared sub-structure so I imagine that could be more efficient.

michalhabera: I am now applying hash at each node and storing hashed byte-strings for the `cache`, this seems to work very well (negligible perf cost).

michalhabera: I am happy with this, any comments @wence- ?

ufl/algorithms/signature.py
```diff
@@ -93,21 +93,22 @@ def compute_terminal_hashdata(expressions, renumbering):
...
-            data = expr._ufl_typecode_  # TODO: Use expr._ufl_signature_data_()? More extensible, but more overhead.
-        expression_hashdata.append(data)
-    # Oneliner: TODO: Benchmark, maybe use a generator?
-    # expression_hashdata = [(terminal_hashdata[expr] if expr._ufl_is_terminal_ else expr._ufl_typecode_)
-    #                       for expr in pre_traversal(expression)]
-    return expression_hashdata
+            data = [expr._ufl_typecode_]
+
+            for op in expr.ufl_operands:
+                data += [cache[op]]
```
wence-, line 109: So I know that previously doing `unique_pre_traversal` you were worried that this one-level expansion was potentially problematic: is that no longer the case with `unique_post_traversal`?

ufl/algorithms/signature.py
```diff
@@ -121,8 +122,7 @@ def compute_expression_signature(expr, renumbering):  # FIXME: Fix callers
 
     # Pass it through a seriously overkill hashing algorithm
     # (should we use sha1 instead?)
-    data = str(expression_hashdata).encode("utf-8")
-    return hashlib.sha512(data).hexdigest()
+    return hashlib.sha512(expression_hashdata).hexdigest()
```
wence-, line 125: Since the final return value is now already hashed do you need to do so again?

ufl/algorithms/signature.py
```diff
@@ -93,21 +93,22 @@ def compute_terminal_hashdata(expressions, renumbering):
...
-        expression_hashdata.append(data)
-    # Oneliner: TODO: Benchmark, maybe use a generator?
-    # expression_hashdata = [(terminal_hashdata[expr] if expr._ufl_is_terminal_ else expr._ufl_typecode_)
-    #                       for expr in pre_traversal(expression)]
-    return expression_hashdata
+            data = [expr._ufl_typecode_]
+
+            for op in expr.ufl_operands:
+                data += [cache[op]]
+        cache[expr] = hashlib.sha512(str(data).encode("utf-8")).digest()
```
wence-, line 110: Is it necessary to hash every operand, or is it sufficient to strify? I guess if the cost of this approach is basically zero then it doesn't really matter too much.

ufl/algorithms/signature.py
michalhabera: No longer the case, since unique post traversal produces full expansion. E.g. the above example `c2*c2+c1` just becomes `hash(sum(hash(product(hash(c2), hash(c2))), hash(c1)))`.

ufl/algorithms/signature.py
michalhabera: The cost of hashing is not significant. It is necessary to hash the tuple or operands, since their stringification would explode the memory (my crazy UFL form got always `Killed`). So I need to compress the already visited expression subtree with hashlib.

wence- (approved):
Thanks!

ufl/algorithms/signature.py
wence-: Thanks, makes sense.

ufl/algorithms/signature.py
wence-: 👍

ufl/algorithms/signature.py
wence-: good to go now!

PR #95, opened 2022-04-07
[feature][ElementTypes] Added 4d element types

dr-robertk:
This PR adds 4d element types to cell.py and elementlist.py.
While for the 4d cube there seems to be only one name, tesseract, for the 4d simplex it's not so clear. 
Here I chose pentatope following the paper https://www.sciencedirect.com/science/article/abs/pii/S0168927421000994.


wence- (approved):
I think I am happy with this (and the names).

dr-robertk:
Thanks. I'm not sure if there are other places in UFL that need adjustment. This patch worked for me though.

dr-robertk:
Thanks for taking care of it.

mscroggs:
Merging this into a side branch so I can correctly merge the changes to cell on main into this PR

PR #107, opened 2022-06-29
trimmed serendipity elements

rckirby:
We've added (and implemented over in Firedrake land) the trimmed serendipity spaces per
Gillette/Kloefkorn/Sanders (http://www.numdam.org/item/10.5802/smai-jcm.41.pdf).

This just updates the list of registered elements.


jmv2009:
Not sure if it's going to work though in Fenics without dual basis.

rckirby:
Fenics doesn’t support Bell or Argyris, either, but they still appear in the UFL registry.  The element list doesn’t require a UFL user to implement the entire set.

dham (approved):

ufl/finiteelement/elementlist.py
```diff
@@ -200,6 +200,10 @@ def show_elements():
                   ("quadrilateral",))
 register_element2("BDMCF", 1, HDiv, "contravariant Piola", (1, None),
                   ("quadrilateral",))
+register_element2("SminusE", 1, HCurl, "covariant Piola", (1, None), cubes[1:3])
+register_element2("SminusF", 1, HDiv, "contravariant Piola", (1, None), cubes[1:2])
```
dham, line 204: Is there a reason that `SminusF` is not allowed on Hexahedra?

dham (changes requested):
See the issue with `SminusF` on Hex

ufl/finiteelement/elementlist.py
rckirby: After digging a bit, it looks like we have two attempts at this (from Cyrus + Justin) that have come in under different names and with slightly different implementations?  I think Cyrus put in SminusE/F and Justin put in SminusDiv/Curl.  The latter supports hexes in practice, but I don't see what the former shouldn't in theory (currently, it's disallowed in FIAT on hexes).

rckirby: If these are the same math object, perhaps we should merge them/make one an alias and wipe out the FI(n)AT implementation of the less complete one?

dham:
superceded by #195

PR #109, opened 2022-07-27
Add `is_fully_continuous` function to finite element

mscroggs:
.. and remove a check on the family string name elsewhere.

This allows anyone writing a custom subclass of `finiteelementbase` to control which elements are continuous without being limited to three strings for family type

### DIFF
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -131,9 +131,7 @@ class RestrictionPropagator(MultiFunction):
         e = o.ufl_element()
         d = e.degree()
         f = e.family()
-        # TODO: Move this choice to the element class?
-        continuous_families = ["Lagrange", "Q", "S"]
-        if (f in continuous_families and d > 0) or f == "Real":
+        if e.is_fully_continuous():
             # If the coefficient _value_ is _fully_ continuous
             return self._default_restricted(o)  # Must still be computed from one of the sides, we just don't care which
         else:
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -129,7 +129,7 @@ class RestrictionPropagator(MultiFunction):
     def coefficient(self, o):
         "Allow coefficients to be unrestricted (apply default if so) if the values are fully continuous across the facet."
         e = o.ufl_element()
-        if e.is_fully_continuous():
+        if e.is_continuous():
             # If the coefficient _value_ is _fully_ continuous
             return self._default_restricted(o)  # Must still be computed from one of the sides, we just don't care which
         else:
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -129,8 +129,6 @@ class RestrictionPropagator(MultiFunction):
     def coefficient(self, o):
         "Allow coefficients to be unrestricted (apply default if so) if the values are fully continuous across the facet."
         e = o.ufl_element()
-        d = e.degree()
-        f = e.family()
         if e.is_fully_continuous():
             # If the coefficient _value_ is _fully_ continuous
             return self._default_restricted(o)  # Must still be computed from one of the sides, we just don't care which
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -129,6 +129,8 @@ class RestrictionPropagator(MultiFunction):
     def coefficient(self, o):
         "Allow coefficients to be unrestricted (apply default if so) if the values are fully continuous across the facet."
         e = o.ufl_element()
+        print(e)
+        print(e.is_fully_continuous())
         if e.is_fully_continuous():
             # If the coefficient _value_ is _fully_ continuous
             return self._default_restricted(o)  # Must still be computed from one of the sides, we just don't care which
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -129,8 +129,6 @@ class RestrictionPropagator(MultiFunction):
     def coefficient(self, o):
         "Allow coefficients to be unrestricted (apply default if so) if the values are fully continuous across the facet."
         e = o.ufl_element()
-        print(e)
-        print(e.is_fully_continuous())
         if e.is_fully_continuous():
             # If the coefficient _value_ is _fully_ continuous
             return self._default_restricted(o)  # Must still be computed from one of the sides, we just don't care which
--- a/ufl/finiteelement/finiteelement.py
+++ b/ufl/finiteelement/finiteelement.py
@@ -242,4 +242,4 @@ class FiniteElement(FiniteElementBase):
         """Return true if the values of this element's basis functions are continuous between elements."""
         if self.family() == "Real":
             return True
-        return self._sobolev_space in H1
+        return self in H1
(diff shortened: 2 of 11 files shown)


ufl/algorithms/apply_restrictions.py
```diff
@@ -129,11 +129,7 @@ def reference_value(self, o):
     def coefficient(self, o):
         "Allow coefficients to be unrestricted (apply default if so) if the values are fully continuous across the facet."
         e = o.ufl_element()
-        d = e.degree()
-        f = e.family()
-        # TODO: Move this choice to the element class?
-        continuous_families = ["Lagrange", "Q", "S"]
-        if (f in continuous_families and d > 0) or f == "Real":
+        if e.is_fully_continuous():
```
dham, line 132: Is this really the right fix? ufl already tracks the Sobolev space embedding of finite elements so I think this could be written `e in ufl.sobolevspace.H1` without adding a sui generis method to elements that only deals with the particular case of fully continuous.

wence-: CR is in h1 (noncorming) according to ufl, but not fully continuous I think

mscroggs: Looks to me like UFL has CR in L2. I'll try making this function check if the element is in H1 or H2 and see if all the tests still pass...

mscroggs: The tests still pass 🥳

ufl/finiteelement/finiteelement.py
```diff
@@ -236,3 +237,9 @@ def __getnewargs__(self):
                 None,
                 self.quadrature_scheme(),
                 self.variant())
+
+    def is_fully_continuous(self):
+        """Return true if the values of this element's basis functions are continuous between elements."""
+        if self.family() == "Real":
+            return True
+        return self._sobolev_space in [H1, H2]
```
dham, line 245: The Sobolev spaces support `in` for elements and also know about which spaces are subspaces of each other. I therefore think `self in H1` would work here.

garth-wells:
Is `is_fully_continuous` a pleonasm? Would `is_continuous` suffice?

ufl/finiteelement/finiteelement.py
```diff
@@ -236,3 +237,9 @@ def __getnewargs__(self):
                 None,
                 self.quadrature_scheme(),
                 self.variant())
+
+    def is_fully_continuous(self):
+        """Return true if the values of this element's basis functions are continuous between elements."""
+        if self.family() == "Real":
+            return True
+        return self in H1
```
wence-, line 245: Probably "Real" finite elements should report that they are in $C^\infty$ and then this special-casing should not be necessary.

ufl/finiteelement/mixedelement.py
```diff
@@ -243,6 +243,13 @@ def shortstr(self):
         tmp = ", ".join(element.shortstr() for element in self._sub_elements)
         return "Mixed<" + tmp + ">"
 
+    def is_fully_continuous(self):
+        """Return true if the values of this element's basis functions are continuous between elements."""
+        for e in self._sub_elements:
+            if not e.is_fully_continuous():
+                return False
+        return True
```
wence-, line 251: `return all(e.is_fully_continuous() for e in e._sub_elements)`

ufl/finiteelement/restrictedelement.py
```diff
@@ -96,3 +96,7 @@ def restricted_sub_elements(self):
         #        w.r.t. different sub_elements meanings.
         "Return list of restricted sub elements."
         return (self._element,)
+
+    def is_fully_continuous(self):
+        """Return true if the values of this element's basis functions are continuous between elements."""
+        return self._element.is_fully_continuous()
```
wence-, line 102: Is this right? If I take the restriction of an $H^1$ space to the interior, is the resulting space continuous, I would say no, there's nothing pinning down the continuity on the faces and vertices any more.

ufl/finiteelement/finiteelementbase.py
```diff
@@ -213,3 +213,7 @@ def __getitem__(self, index):
             from ufl.finiteelement import RestrictedElement
             return RestrictedElement(self, index)
         return NotImplemented
+
+    def is_fully_continuous(self):
+        """Return true if the values of this element's basis functions are continuous between elements."""
+        raise NotImplementedError()
```
wence-, line 219: Why not implement the method on this base class and then the only special case is for mixed element?

mscroggs:
> Is `is_fully_continuous` a pleonasm? Would `is_continuous` suffice?

The "fully" makes it clear that it's more than just (eg) continuity in the normal or tangential directions. Also helps to make it clear that this isn't the same as the difference between "discontinuous Raviart-Thomas" and "continuous Raviart-Thomas".

ufl/finiteelement/finiteelementbase.py
mscroggs: The base class doesn't have a sobolev space, so it didn't look like `self in H1` would work in the base class

ufl/finiteelement/restrictedelement.py
mscroggs: I've changed this to return False

ufl/finiteelement/finiteelement.py
mscroggs: I agree, but as there's no Cinfinity space in sobolevspace.py yet, I think this might be one for a future PR. I'll open an issue for it.

mscroggs: #114

garth-wells:
> The "fully" makes it clear that it's more than just (eg) continuity in the normal or tangential directions. Also helps to make it clear that this isn't the same as the difference between "discontinuous Raviart-Thomas" and "continuous Raviart-Thomas".

What 'fully' adds wasn't clear to me. It seems obvious to me that an element with only some continuous components isn't continuous. I think something either is or isn't continuous.

On @dham's point: isn't just testing for H1 adequate? Is this PR just for handling 'constants'?

ufl/finiteelement/finiteelementbase.py
dham: The base class doesn't need to have a Sobolev space. Python instance methods are called with the concrete object as `self`, so the object that it gets called on ought to have one.

ufl/finiteelement/finiteelement.py
dham: You could get rid of this special case by just shifting "Real" into "H2", which is more accurate than the current situation with it being in "L2".

dham:
> > The "fully" makes it clear that it's more than just (eg) continuity in the normal or tangential directions. Also helps to make it clear that this isn't the same as the difference between "discontinuous Raviart-Thomas" and "continuous Raviart-Thomas".
> 
> What 'fully' adds wasn't clear to me. It seems obvious to me that an element with only some continuous components isn't continuous. I think something either is or isn't continuous.
> 
> On @dham's point: isn't just testing for H1 adequate? Is this PR just for handling 'constants'?

I agree. I think this PR is sort of the wrong design. It's really asking if the element is in `H1` and the clean thing to do would be to fix the cases where elements don't correctly report their Sobolev Space and just use that. This also avoids the nomenclature disagreement about "fully" because it's clear that H1 means continuous in all components.

dham:
OK so I allowed myself to get distracted by this and extended SobolevSpace to deal with this situation. I think it's a nicer solution. It's over at #117

PR #110, opened 2022-07-27
Change DOLFINx integration test image

jhale:


mscroggs:
You need to set the compilers in the the ffcx tests workflow to make Basix install (as I've done here in #109: https://github.com/FEniCS/ufl/blob/845ed3de0b138e9bdc5eeba6482f950fd1f3795f/.github/workflows/fenicsx-tests.yml#L14-L16)

PR #111, opened 2022-08-04
Sanitise unicode input

dham:
Add the Trimmed Serendipity elements to the element list.


dham:
The commit history here is a mess, but. since it's only 4 lines, the easiest and best way to deal with that is by squash on merge.

garth-wells (approved):

PR #113, opened 2022-08-10
Fix signature for forms with identitcal constants and identical meshes (up to ufl_id)

jorgensd:
Reported at: https://fenicsproject.discourse.group/t/constant-in-dolfinx-and-ffcx-compilation/8968?u=dokken

Minimal example failing on main:
```python

import ufl

mesh = ufl.Mesh(ufl.VectorElement("Lagrange", ufl.triangle, 1))
c = ufl.Constant(mesh)
d = ufl.Constant(mesh)
a = c * ufl.dx
b = d * ufl.dx
assert(a.signature() == b.signature())
```

Constants now has the same behavior as coefficients:
```python
V = ufl.FunctionSpace(mesh, ufl.FiniteElement("CG", mesh.ufl_cell(), 1))
u = ufl.Coefficient(V)
v = ufl.TestFunction(V)
w = ufl.Coefficient(V)

a = ufl.inner(u, v) * ufl.dx
d = ufl.inner(w, v) * ufl.dx
assert(a.signature() == d.signature())
```

This PR now also supports same signature for different meshes (with same coordinate element), 
i.e.
```python
mesh = ufl.Mesh(ufl.VectorElement("Lagrange", ufl.triangle, 1))
mesh2 = ufl.Mesh(ufl.VectorElement("Lagrange", ufl.triangle, 1))
c = ufl.Constant(mesh)
d = ufl.Constant(mesh2)
a = c * ufl.dx
b = d * ufl.dx
assert(a.signature() == b.signature())
```

### DIFF
--- a/ufl/constant.py
+++ b/ufl/constant.py
@@ -69,7 +69,8 @@ class Constant(Terminal):
     def _ufl_signature_data_(self, renumbering):
         "Signature data for constant depends on renumbering"
         return "Constant({}, {}, {})".format(
-            repr(self._ufl_domain), repr(self._ufl_shape), repr(renumbering[self]))
+            self._ufl_domain._ufl_signature_data_(renumbering), repr(self._ufl_shape),
+            repr(renumbering[self]))
 
 
 def VectorConstant(domain, count=None):
--- a/ufl/constant.py
+++ b/ufl/constant.py
@@ -66,6 +66,11 @@ class Constant(Terminal):
                 self._ufl_domain == other._ufl_domain and
                 self._ufl_shape == self._ufl_shape)
 
+    def _ufl_signature_data_(self, renumbering):
+        "Signature data for constant depends on renumbering"
+        return "Constant({}, {}, {})".format(
+            repr(self._ufl_domain), repr(self._ufl_shape), repr(renumbering[self]))
+
 
 def VectorConstant(domain, count=None):
     domain = as_domain(domain)
--- a/ufl/form.py
+++ b/ufl/form.py
@@ -81,6 +81,7 @@ class Form(object):
         "_arguments",
         "_coefficients",
         "_coefficient_numbering",
+        "_constant_numbering",
         "_constants",
         "_hash",
         "_signature",
@@ -111,9 +112,12 @@ class Form(object):
         self._arguments = None
         self._coefficients = None
         self._coefficient_numbering = None
+        self._constant_numbering = None
 
         from ufl.algorithms.analysis import extract_constants
         self._constants = extract_constants(self)
+        self._constant_numbering = dict(
+            (c, i) for i, c in enumerate(self._constants))
 
         # Internal variables for caching of hash and signature after
         # first request
@@ -236,6 +240,11 @@ class Form(object):
 
     def constants(self):
         return self._constants
+
+    def constant_numbering(self):
+        """Return a contiguous numbering of constants in a mapping
+        ``{constant:number}``."""
+        return self._constant_numbering
 
     def signature(self):
         "Signature for use with jit cache (independent of incidental numbering of indices etc.)"
@@ -458,9 +467,11 @@ class Form(object):
         # Include integration domains and coefficients in renumbering
         dn = self.domain_numbering()
         cn = self.coefficient_numbering()
+        cnstn = self.constant_numbering()
         renumbering = {}
         renumbering.update(dn)
         renumbering.update(cn)
+        renumbering.update(cnstn)
 
         # Add domains of coefficients, these may include domains not
         # among integration domains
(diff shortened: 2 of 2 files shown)


jorgensd:
@dham This change should not affect firedrake, as it is using `ufl.Coefficient` for constants (https://github.com/firedrakeproject/firedrake/blob/master/firedrake/constant.py#L26)
which is the same as legacy dolfin does.

dham:
> @dham This change should not affect firedrake, as it is using `ufl.Coefficient` for constants (https://github.com/firedrakeproject/firedrake/blob/master/firedrake/constant.py#L26)
> which is the same as legacy dolfin does.

I'm sure that's right. I've just set https://github.com/firedrakeproject/firedrake/pull/2521 running to be sure. If that passes then I'm happy.

dham (approved):

thisch:
There still seems to be problems in builds with a complex `PetscScalar`. 

I've tried running the demo from the linked discourse issue. What I observe is that when I pass a `float` to `fem.Constant` it fails to compile the form, whereas if I cast the number to a complex, it works.

However, for this part of the MWE

```py
# this does not recompile (but I need the former)
cnst = Constant(mesh,1.) 
for c in np.linspace(0,.2,10):
    cnst.value = complex(c, 0)
    f = assemble_scalar(form(cnst * dx(domain=mesh),jit_params=jit_params))
    print(f)
```
It doesnt even work when I pass a complex to `cnst.value`.

```
Traceback (most recent call last):
  File "/home/thomas/gitrepos/fenics_laser_states_2022/demos/fenics_qa_ffc_demo.py", line 24, in <module>
    cnst.value = complex(c, 0)
  File "/home/thomas/miniconda/envs/fenicsx/lib/python3.10/site-packages/dolfinx/fem/function.py", line 63, in value
    np.copyto(self._cpp_object.value, np.asarray(v))
  File "<__array_function__ internals>", line 180, in copyto
TypeError: Cannot cast scalar from dtype('complex128') to dtype('float64') according to the rule 'same_kind'
```

jorgensd:
> There still seems to be problems in builds with a complex `PetscScalar`.
> 
> I've tried running the demo from the linked discourse issue. What I observe is that when I pass a `float` to `fem.Constant` it fails to compile the form, whereas if I cast the number to a complex, it works.
> 
> However, for this part of the MWE
> 
> ```python
> # this does not recompile (but I need the former)
> cnst = Constant(mesh,1.) 
> for c in np.linspace(0,.2,10):
>     cnst.value = complex(c, 0)
>     f = assemble_scalar(form(cnst * dx(domain=mesh),jit_params=jit_params))
>     print(f)
> ```
> 
> It doesnt even work when I pass a complex to `cnst.value`.
> 
> ```
> Traceback (most recent call last):
>   File "/home/thomas/gitrepos/fenics_laser_states_2022/demos/fenics_qa_ffc_demo.py", line 24, in <module>
>     cnst.value = complex(c, 0)
>   File "/home/thomas/miniconda/envs/fenicsx/lib/python3.10/site-packages/dolfinx/fem/function.py", line 63, in value
>     np.copyto(self._cpp_object.value, np.asarray(v))
>   File "<__array_function__ internals>", line 180, in copyto
> TypeError: Cannot cast scalar from dtype('complex128') to dtype('float64') according to the rule 'same_kind'
> ```

As seen in the source code of DOLFINx: https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/function.py#L32-L57
the constant is created wrt. to the `dtype` of the original input. I.e. for complex builds you should send in the right dtype at compilation time, see:
```python
from petsc4py import PETSc
from dolfinx.fem import (Constant, assemble_scalar, form)
from dolfinx.mesh import create_unit_square
from mpi4py import MPI
from ufl import dx
import numpy as np
from pathlib import Path
import shutil

mesh = create_unit_square(MPI.COMM_WORLD, 1, 1)

cache_dir = f"{str(Path.cwd())}/cache"
shutil.rmtree(cache_dir, ignore_errors=True)
jit_options = {"cache_dir": cache_dir, "cffi_verbose": True}
# this recompiles the form at each iterations (different behavior wrt dolfin-old)
for c in np.linspace(0, 1, 10):
    cnst = Constant(mesh, PETSc.ScalarType(c))
    f = assemble_scalar(form(cnst * dx(domain=mesh), jit_options=jit_options))
    print(f)

# this does not recompile (but I need the former)
cnst = Constant(mesh, PETSc.ScalarType(1.))
for c in np.linspace(0, .2, 10):
    cnst.value = c
    f = assemble_scalar(form(cnst * dx(domain=mesh), jit_options=jit_options))
    print(f)
```
for an example compatible with `dolfinx/dolfinx:nightly` built the 20th of October. The only difference for the 0.5.1 release is the change of `jit_params` to `jit_options`.

thisch:
Thx @jorgensd, this works!

PR #117, opened 2022-08-12
H inf sobolev space

dham:
Define an `HInf` Sobolev space to contain the "R" space.

Cause apply_restrictions to use Sobolev space rather than a list of elements. This is an alternative to #109 .

Also clean up a few Sobolev space loose ends which appear to have drifted in along the way.

### DIFF
--- a/ufl/sobolevspace.py
+++ b/ufl/sobolevspace.py
@@ -15,6 +15,7 @@ symbolic reasoning about the spaces in which finite elements lie."""
 # Modified by Thomas Gibson 2017
 
 from functools import total_ordering
+from math import inf, isinf
 
 
 @total_ordering
@@ -39,6 +40,7 @@ class SobolevSpace(object):
             "L2": 0,
             "H1": 1,
             "H2": 2,
+            "HInf": inf,
             # Order for the elements below is taken from
             # its parent Sobolev space
             "HDiv": 0,
@@ -114,11 +116,8 @@ class DirectionalSobolevSpace(SobolevSpace):
                      the position denotes in what spatial variable the
                      smoothness requirement is enforced.
         """
-        assert all(
-            isinstance(x, int) for x in orders), ("Order must be an integer.")
-        assert all(
-            x < 3
-            for x in orders), ("Not implemented for orders greater than 2")
+        assert all(isinstance(x, int) or isinf(x) for x in orders), \
+            ("Order must be an integer or infinity.")
         name = "DirectionalH"
         parents = [L2]
         super(DirectionalSobolevSpace, self).__init__(name, parents)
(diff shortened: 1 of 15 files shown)


dham:
Firedrake tests for this running at https://github.com/firedrakeproject/firedrake/pull/2522

mscroggs (approved):

ufl/sobolevspace.py
```diff
@@ -179,5 +178,6 @@ def __str__(self):
 HCurl = SobolevSpace("HCurl", [L2])
 H1 = SobolevSpace("H1", [HDiv, HCurl, L2])
 H2 = SobolevSpace("H2", [H1, HDiv, HCurl, L2])
+HInf = SobolevSpace("HInf", [H2, H1, HDiv, HCurl, L2])
```
wence-, line 181: While we're here, can we update the parent list for all of these to include the `HEin` and `HDivDiv` spaces below, I think that they have stronger continuity requirements than `HCurl` and `HDiv` respectively. Although maybe one doesn't want that since really `HDivDiv in Vector(HDiv)` since the former is tensor fields and the latter vector fields.

mscroggs:
DOLFINx tests are failing with "AttributeError: 'VectorElement' object has no attribute 'sobolev_space'". I guess either VectorElements and other elements need to be given Sobolev spaces or we do something like #109

dham:
> DOLFINx tests are failing with "AttributeError: 'VectorElement' object has no attribute 'sobolev_space'". I guess VectorElements and other elements need to be given Sobolev spaces (or we do something like #109, but I agree that this approach is probably better)

It's 2030 on Friday evening here so I'm not going to get to this until Monday, but VectorElement should inherit the Sobolev space of it's scalar element. TensorElement should do the same. Oh, and BrokenElement should always be L2, no matter what the space of the original element (that being the point of BrokenElement).

mscroggs:
> It's 2030 on Friday evening here so I'm not going to get to this until Monday, but VectorElement should inherit the Sobolev space of it's scalar element. TensorElement should do the same. Oh, and BrokenElement should always be L2, no matter what the space of the original element (that being the point of BrokenElement).

I've just started working on this in #118 and I'm relieved that these are exactly what I just implemented

mscroggs:
The DOLFINx tests pass in #118. @dham are you happy for me to merge that branch into this one?

ufl/sobolevspace.py
wence-: Let's pass on this one.

wence- (approved):
I think this looks good

ufl/sobolevspace.py
mscroggs: Do you want to open an issue for this?

garth-wells (approved):
Note for a future UFL: we need to start from 'domains', on which spaces are defined, to make proper sense of function spaces.

PR #118, opened 2022-08-12
Add sobolev_space method to base finite element class

mscroggs:
Added sobolev_space method to base finite element class

Made base finite element class into an abc abstract class, so that elements cannot be created without methods such as sobolev_space being implemented

This builds on #117, and is currently set to merge into that branch so it can be part of that PR.

### DIFF
--- a/ufl/finiteelement/mixedelement.py
+++ b/ufl/finiteelement/mixedelement.py
@@ -124,7 +124,7 @@ class MixedElement(FiniteElementBase):
         return sm or EmptyDict
 
     def sobolev_space(self):
-        return min(e.sobolev_space() for e in self._sub_elements)
+        return max(e.sobolev_space() for e in self._sub_elements)
 
     def mapping(self):
         if all(e.mapping() == "identity" for e in self._sub_elements):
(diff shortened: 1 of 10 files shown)


ufl/finiteelement/mixedelement.py
```diff
@@ -125,6 +123,9 @@ def symmetry(self):
             error("Size mismatch in symmetry algorithm.")
         return sm or EmptyDict
 
+    def sobolev_space(self):
+        return min(e.sobolev_space() for e in self._sub_elements)
+
```
dham, line 128: Is this the right thing? Suppose I have `CG1 x DG0`. I think this will say that the whole thing is in `H1`

dham (changes requested):

ufl/finiteelement/mixedelement.py
mscroggs: Yes, looks like max might be what we actually want here?

dham: I think that's a safe conservative option. It might disallow implicit restrictions in some cases where they would actually work, but that can be dealt with in a later pass if someone cares enough.

dham (approved):

PR #119, opened 2022-08-15
Override is_cellwise_constant for CellNormal

michalhabera:
Following MWE produces `True`, 
```python
import ufl

coord_el = ufl.VectorElement("CG", ufl.triangle, 2, dim=3)
mesh = ufl.Mesh(coord_el)

n = ufl.CellNormal(mesh)
print(n.is_cellwise_constant())
```
which results in terms like `grad(CellNormal)` being incorrectly simplified to `Zero`.


dham (approved):
Looks correct to me.

garth-wells (approved):

PR #120, opened 2022-09-06
Allow multiple subdomain data

jpdean:
Currently, UFL checks that all integrals of a particular type in a form have the same subdomain data (using the UFL ID). I'm not sure this should be UFL's job. Instead,  UFL should probably just collect the data into a list and let the finite element library (e.g. dolfinx) either check the data is the same or make use of the additional data. This PR addresses this.

This PR requires https://github.com/FEniCS/dolfinx/pull/2369

### DIFF
--- a/.github/workflows/fenicsx-tests.yml
+++ b/.github/workflows/fenicsx-tests.yml
@@ -86,7 +86,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: jpdean/multiple_subdomain_data
+          ref: main
       - name: Install DOLFINx
         run: |
           cmake -G Ninja -DCMAKE_BUILD_TYPE=Developer -B build -S dolfinx/cpp/
--- a/.github/workflows/fenicsx-tests.yml
+++ b/.github/workflows/fenicsx-tests.yml
@@ -86,7 +86,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: jpdean/manual_integration_domains_2
+          ref: jpdean/multiple_subdomain_data
       - name: Install DOLFINx
         run: |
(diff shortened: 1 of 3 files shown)


chrisrichardson (approved):

.github/workflows/fenicsx-tests.yml
```diff
@@ -86,7 +86,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: main
+          ref: jpdean/manual_integration_domains_2
```
wence-, line 89: don't forget to revert this before merge

chrisrichardson: Is this needed to pass the integration test?

jpdean: Yes, but only because dolfinx main doesn't expect a list to be returned. I'll make a new branch of dolfinx now with a small change to account for this.

test/test_measures.py
```diff
@@ -185,23 +185,33 @@ def test_foo():
...
     assert domain.ufl_cargo() == mymesh
-    assert Mx.subdomain_data()[mydomain]["cell"] == cell_domains
+    assert(len(Mx.subdomain_data()[mydomain]["cell"]) == 1)
+    assert(Mx.subdomain_data()[mydomain]["cell"][0]) == cell_domains
 
     domain, = Ms.ufl_domains()
     assert domain.ufl_cargo() == mymesh
-    assert Ms.subdomain_data()[mydomain][
-        "exterior_facet"] == exterior_facet_domains
+    assert(len(Ms.subdomain_data()[mydomain][
```
garth-wells, line 193: I think these asserts would be much more readable without the line breaks.

garth-wells (approved):

PR #121, opened 2022-09-15
Add variant to EnrichedElement, TensorProductElement, etc

pbrubeck:
This PR lets composite elements return the `variant` property of their sub-elements. When the sub-elements are of different variants, `variant()` returns `None`.

### DIFF
--- a/ufl/finiteelement/enrichedelement.py
+++ b/ufl/finiteelement/enrichedelement.py
@@ -82,6 +82,13 @@ class EnrichedElementBase(FiniteElementBase):
             sobolev_space, = intersect
             return sobolev_space
 
+    def variant(self):
+        elements = [e for e in self._elements]
+        if all(e.variant() == elements[0].variant()
+               for e in elements):
+            return elements[0].variant()
+        return None
+
     def reconstruct(self, **kwargs):
         return type(self)(*[e.reconstruct(**kwargs) for e in self._elements])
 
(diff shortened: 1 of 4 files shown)


ufl/finiteelement/enrichedelement.py
```diff
@@ -85,6 +85,13 @@ def sobolev_space(self):
             sobolev_space, = intersect
             return sobolev_space
 
+    def variant(self):
+        elements = [e for e in self._elements]
```
wence-, line 89: aka:

```
try:
    variant, = {e.variant() for e in self._elements}
    return variant
except ValueError:
    return None
```

ufl/finiteelement/elementlist.py
```diff
@@ -200,6 +200,10 @@ def show_elements():
                   ("quadrilateral",))
 register_element2("BDMCF", 1, HDiv, "contravariant Piola", (1, None),
                   ("quadrilateral",))
+register_element2("SminusE", 1, HCurl, "covariant Piola", (1, None), cubes[1:3])
+register_element2("SminusF", 1, HDiv, "contravariant Piola", (1, None), cubes[1:2])
+register_element2("SminusDiv", 1, HDiv, "contravariant Piola", (1, None), cubes[1:3])
+register_element2("SminusCurl", 1, HCurl, "covariant Piola", (1, None), cubes[1:3])
```
wence-, line 206: Why are these part of this change?

pbrubeck: I think these "Sminus" elements aren't in the FEniCS UFL yet? Anyways I have cleaned the commit history.

PR #122, opened 2022-09-26
Replace `not isinstance(element, MixedElement)` with `element.num_sub_elements == 0`

mscroggs:
This allows me to use control the behaviour of (eg) `TrialFunctions` for custom elements.

~As a side effect, this may change the result of calling `TrailFunctions` on `VectorElement`s and `TensorElement`s, although this new behaviour is probably what I'd have expected it to do.~ Edit: VectorElement inherits from MixedElement, so the behaviour probably hasn't changed


dham:
I think this looks fine. I have started https://github.com/firedrakeproject/firedrake/pull/2568 running just to make sure.

dham:
> I think this looks fine. I have started [firedrakeproject/firedrake#2568](https://github.com/firedrakeproject/firedrake/pull/2568) running just to make sure.

So it turns out that the sobolev spaces change (which hadn't yet been merged to Firedrake) failed to set the `sobolev_space` method on `BrokenElement` which causes this to fail Firedrake tests. I am testing a fix.

PR #126, opened 2022-10-11
Implement sobolev_space() for BrokenElement and WithMapping

pbrubeck:
Since #117, `sobolev_space()` is an abstract method, and a couple elements were not implementing this method.

### DIFF
--- a/ufl/finiteelement/hdivcurl.py
+++ b/ufl/finiteelement/hdivcurl.py
@@ -81,7 +81,7 @@ class HCurlElement(FiniteElementBase):
         return "covariant Piola"
 
     def sobolev_space(self):
-        "Return the underlying Sobolev space."
+        """Return the underlying Sobolev space."""
         return HCurl
 
     def reconstruct(self, **kwargs):
@@ -94,7 +94,7 @@ class HCurlElement(FiniteElementBase):
         return f"HCurlElement({repr(self._element)})"
 
     def shortstr(self):
-        "Format as string for pretty printing."
+        """Format as string for pretty printing."""
         return f"HCurlElement({self._element.shortstr()})"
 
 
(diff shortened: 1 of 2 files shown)


ufl/finiteelement/hdivcurl.py
```diff
@@ -143,6 +143,10 @@ def reference_value_shape(self):
     def mapping(self):
         return self._mapping
 
+    def sobolev_space(self):
+        """Return the underlying Sobolev space."""
+        return self.wrapee.sobolev_space()
```
dham, line 148: I think this one is wrong. If you change the pullback then in general you're going to change the continuity of the element and hence the Sobolev space in which it is embedded. I think what we need to do here is give the `WithMapping` constructor a `sobolev_space=L2` optional argument, and set `self._sobolev_space` to that and finally return that attribute here.

The alternative is to set the Sobolev space according to the mapping requested, but it's not obvious to me that that's safe.

dham (changes requested):
Thanks for this. I just noticed this issue over on #122 so I'm very happy you've fixed it. I'm just not sure that what is proposed for WithMapping is correct.

ufl/finiteelement/hdivcurl.py
wence-: I think the way that sobolev spaces are used, `L2` is always safe? UFL uses it to determine if restricted elements are continuous across faces, not for determining which pullback to apply. So would agree that defaulting to `L2` is a good choice.

pbrubeck: I defaulted to L2, but in case WithMapping is a no-op we may return `self.wrapee.sobolev_space()`

dham (approved):

PR #131, opened 2022-11-08
Remove deprecated things

mscroggs:
Delete functions that have deprecation warnings


ufl/argument.py
```diff
@@ -74,14 +74,14 @@ def ufl_function_space(self):
 
     def ufl_domain(self):
         "Deprecated, please use .ufl_function_space().ufl_domain() instead."
-        # TODO: deprecate("Argument.ufl_domain() is deprecated, please
-        # use .ufl_function_space().ufl_domain() instead.")
+        deprecate("Argument.ufl_domain() is deprecated, please "
```
jorgensd, line 77: As mentioned in https://github.com/FEniCS/ufl/pull/80#pullrequestreview-853581102 it is unclear if deprecating this is a step in the right direction

mscroggs: Potentially, this deprecation warning comment should be removed rather than what I did

mscroggs: I've un-deprecated these two functions

chrisrichardson (approved):

PR #137, opened 2022-12-20
Use Python warnings and exceptions instead of custom logging

mscroggs:
~This requires https://github.com/FEniCS/dolfinx/pull/2490 and https://github.com/firedrakeproject/tsfc/pull/285~ (both merged now)


mscroggs:
tsfc is importing GREEN from ufl.log, which this PR removes, so that will need updating

jorgensd (approved):

garth-wells:
`BaseException` isn't very informative for a user. Is there a next step plan for exception types?

mscroggs:
> `BaseException` isn't very informative for a user. Is there a next step plan for exception types?

No, we should work on a plan

garth-wells (approved):

PR #142, opened 2023-01-09
Simplify coeff and constant naming

chrisrichardson:
This fixes a problem with coefficient names in `ffcx`. The current naming is probably for LaTeX compatibility, but I feel this should be a secondary consideration.
See https://github.com/FEniCS/ffcx/issues/551


mscroggs (approved):
Definitely looks like it's there for LaTeX compatibility and I agree with removing it

PR #144, opened 2023-01-15
Fix typo in split

nbouziani:


garth-wells:
Could you add a test? 

Shows inadequacy of the test suite to pick up even basic issues.

garth-wells (approved):

PR #145, opened 2023-01-17
Add tests for legacy DOLFIN

mscroggs:
We should (temporarily) run tests against legacy FFC/DOLFIN so that we know which release is the final release compatible with legacy FEniCS

Temporarily readded (to make compatible with FFC):

- ufl/log.py
- cellname2facetname in ufl/cell.py
- geometric_dimension in core/expr.py

### DIFF
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -84,6 +84,12 @@ num_cell_entities = {"vertex": (1,),
 # Mapping from cell name to topological dimension
 cellname2dim = dict((k, len(v) - 1) for k, v in num_cell_entities.items())
 
+# TODO: Remove this dict
+cellname2facetname = {"interval": "vertex",
+                      "triangle": "interval",
+                      "quadrilateral": "interval",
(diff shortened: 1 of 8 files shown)


ufl/cell.py
```diff
@@ -84,6 +84,13 @@ def __lt__(self, other):
 # Mapping from cell name to topological dimension
 cellname2dim = dict((k, len(v) - 1) for k, v in num_cell_entities.items())
 
+# TODO: Remove this dict
+cellname2facetname = {"interval": "vertex",
```
francesco-ballarin, line 88: I believe that this is needed only in one file in FFC, you may want to add it there rather than adding it back here.
https://github.com/fem-on-colab/fem-on-colab/blob/main/fenics/patches/08-add-cellname2facetname-to-ffc

mscroggs:
Abandoning this, as there are too many things that fail in legacy DOLFIN

PR #146, opened 2023-01-19
Add concurrency to workflows

mscroggs:
This will make GitHub actions cancel workflows that are currently running on a branch if a new commit is pushed

### DIFF
--- a/.github/workflows/fenicsx-tests.yml
+++ b/.github/workflows/fenicsx-tests.yml
@@ -7,6 +7,10 @@ on:
     branches:
       - main
 
+concurrency:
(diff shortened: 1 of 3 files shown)


jorgensd (approved):

mscroggs:
> Is this helpful? Running tests on all changes can be helpful when moving fast to identify which changes cause a failure.

It's definitely helpful for slower workflows, as this will automatically cancel them before you have too many jobs running and everything ends up in a queue. But perhaps these workflows are fast enough that we don't want this here (or maybe we just want it on the fenicsx and tsfc tests).

PR #147, opened 2023-01-26
Use BaseArgument.__eq__ in Argument

michalhabera:
Otherwise, `Argument.__eq__` is inherited from the first parent class, `FormArgument`, which has the equality from `Terminal` which is very inefficient.

I do not know if this is the right thing, but it makes `Argument.__eq__` behave as before the #87 .

Fixes https://github.com/FEniCS/ffcx/issues/552

### DIFF
--- a/ufl/argument.py
+++ b/ufl/argument.py
@@ -152,6 +152,7 @@ class Argument(FormArgument, BaseArgument):
     __getnewargs__ = BaseArgument.__getnewargs__
     __str__ = BaseArgument.__str__
     _ufl_signature_data_ = BaseArgument._ufl_signature_data_
+    __eq__ = BaseArgument.__eq__
 
     def __new__(cls, *args, **kw):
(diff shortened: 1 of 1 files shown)


dham:
This looks right to me. @nbouziani, can we run Firedrake tests on this change just to double check.

michalhabera:
> Do you need to implement `__hash__` too?

I do not think so, `@ufl_type()` attaches default hash as `compute_expr_hash`.

nbouziani:
> This looks right to me. @nbouziani, can we run Firedrake tests on this change just to double check.

This doesn't break anything on our side,

garth-wells:
@michalhabera will this close https://github.com/FEniCS/ffcx/issues/552?

michalhabera:
> @michalhabera will this close [FEniCS/ffcx#552](https://github.com/FEniCS/ffcx/issues/552)?

Yes. I've added this info to the PR desc.

PR #152, opened 2023-02-15
fix TensorProductElement repr

ksagiyam:
A small fix.

### DIFF
--- a/ufl/finiteelement/tensorproductelement.py
+++ b/ufl/finiteelement/tensorproductelement.py
@@ -69,9 +69,7 @@ class TensorProductElement(FiniteElementBase):
         self._cell = cell
 
     def __repr__(self):
-        return "TensorProductElement(" + ", ".join(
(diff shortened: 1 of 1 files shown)


ufl/finiteelement/tensorproductelement.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, *elements, **kwargs):
     def __repr__(self):
         return "TensorProductElement(" + ", ".join(
             repr(e) for e in self._sub_elements
-        ) + f", {repr(self._cell)})"
+        ) + f", cell={repr(self._cell)})"
```
garth-wells, line 74: Could you move the three lines into one? It's very hard to read being split across three lines. I think all on one line will pass the linter.

ksagiyam: Sure. Done!

garth-wells (approved):

Issue #151, opened 2023-02-14
ufl 2023.1 breaks old dolfin


drew-parsons:
The change of API in ufl 2023.1 (2023.1.1.post0) breaks old dolfin.

It also breaks old ffc.  `ufl.log.WARNING` was removed in ufl 2023 and is used by ffc. But ffc can be easily patched to treat any instance of WARNING as either INFO or ERROR.

dolfin cannot be so easily patched. The problem is not only ufl.log.WARNING. For instance demo_poisson.py gives
```
$ python3 demo_poisson.py 
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Traceback (most recent call last):
  File "/projects/fenics/build/tests/dolfin/demo-python/documented/poisson/demo_poisson.py", line 170, in <module>
    solve(a == L, u, bc)
  File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/solving.py", line 233, in solve
    _solve_varproblem(*args, **kwargs)
  File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/solving.py", line 268, in _solve_varproblem
    problem = LinearVariationalProblem(eq.lhs, eq.rhs, u, bcs,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/problem.py", line 58, in __init__
    L = Form(L, form_compiler_parameters=form_compiler_parameters)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/form.py", line 82, in __init__
    self.set_cell_domains(subdomains)
TypeError: set_cell_domains(): incompatible function arguments. The following argument types are supported:
    1. (self: dolfin.cpp.fem.Form, arg0: dolfin.cpp.mesh.MeshFunctionSizet) -> None

Invoked with: <dolfin.fem.form.Form object at 0x7f67280289b0>, [None]
```

[...]

PR #155, opened 2023-03-22
rename namespace from ufl to ufl_legacy

drew-parsons:
ufl 2023.1 diverged sufficiently that it is no longer compatible with old dolfin, and not readily patched to restore compatibility.

The issue is raised at https://github.com/FEniCS/ufl/issues/151 . Discussion at #151 concluded that since ufl is used elsewhere (e.g. firedrake), it is preferable to rename old ufl (2022.2.0) as ufl_legacy, rather than to rename latest ufl as uflx (which was done for ffc->ffcx)

This patch commences the change of namespace from ufl to ufl_legacy, intended for merging into a new ufl_legacy branch based on the 2022.2.0 tag.


drew-parsons:
One question I'm not certain about is the meta information: in setup.cfg should the name be changed from fenics-ufl to fenics-ufl-legacy?  I wasn't sure since it's tied to the version.  We probably want to freeze the major version at 2022 and increment the minor version.  So there won't necessarily be a clash between legacy "fenics-ufl 2022.3.0" and new "fenics-ufl 2023.1.1".

But thinking about it, since (old) dolfin will need to be updated to use the ufl_legacy namespace, it's probably more correct to have it require fenics-ufl-legacy rather than fenics-ufl, so setup.cfg for ufl_legacy should probably be patched this way.

drew-parsons:
I think it makes sense to rename the distribution name as `fenics-ufl-legacy`, in line with renaming the package itself. I think the PR is ready for review now.

drew-parsons:
I missed some ufl components. I'll patch that now

drew-parsons:
I tried restoring the "misplaced" files and being more explicitly in using git mv before changing contents. But then github reported the same state anyway, as if these 4 files (and not the others) were copied and deleted without git mv.  But the successful renaming is seen in 7a708f7 with these 4 files renamed the same way as the others.

So I don't know why the PR summary displays these 4 files differently (algorithms/__init__.py, algorithms/multifunction.py, classes.py,  finiteelement/__init__.py).  Perhaps it's just a glitch in the github web interface.

So I've reset (annulled) those last changes.  The final result is consistent with what it should be. I'm reopening for review.

mscroggs (approved):

PR #162, opened 2023-05-11
fix str

ksagiyam:
This PR fixes `str`.

```
from ufl import *
  

elem = FiniteElement("Q", quadrilateral, 1)
print(elem)
```
currently prints:
```
<None1 on a quadrilateral>
```
instead of:
```
<Q1 on a quadrilateral>
```


ksagiyam:
Sure. Added a test.

dham (approved):

PR #168, opened 2023-06-07
Refactor cells

mscroggs:
Make cells into abstract base classes so cells can be implemented easily outside UFL is desired.

(They may wish do this if, for example, if someone wanted to define a hexagon cell or any other cell not implemented in UFL directly.)

This does not change the current user interface or remove any available functionality, aside from removing a few dicts that no-one should've been using anyway.

This is coupled with https://github.com/FEniCS/ffcx/pull/571

### DIFF
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -190,7 +190,8 @@ class AbstractCell(UFLObject):
 
 class Cell(AbstractCell):
     """Representation of a named finite element cell with known structure."""
-    __slots__ = ("_cellname", "_tdim", "_gdim", "_num_cell_entities", "_sub_entity_types")
+    __slots__ = ("_cellname", "_tdim", "_gdim", "_num_cell_entities", "_sub_entity_types",
+                 "_sub_entities", "_sub_entity_types")
 
     def __init__(self, cellname: str, geometric_dimension: typing.Optional[int] = None):
         if cellname == "vertex":
@@ -220,6 +221,13 @@ class Cell(AbstractCell):
         else:
             raise ValueError(f"Unsupported cell type: {cellname}")
 
+        try:
+            self._sub_entities = [Cell(t, self._gdim) for t in self._sub_entity_types[dim]]
+            self._sub_entity_types = [Cell(t, self._gdim) for t in set(self._sub_entity_types[dim])]
+        except IndexError:
+            self._sub_entities []
+            self._sub_entity_types []
+
         self._cellname = cellname
         self._tdim = len(self._num_cell_entities) - 1
         self._gdim = self._tdim if geometric_dimension is None else geometric_dimension
@@ -241,7 +249,6 @@ class Cell(AbstractCell):
 
     def is_simplex(self) -> bool:
         """Return True if this is a simplex cell."""
-        print(self._cellname)
         return self._cellname in ["vertex", "interval", "triangle", "tetrahedron"]
 
     def has_simplex_facets(self) -> bool:
@@ -257,17 +264,11 @@ class Cell(AbstractCell):
 
     def sub_entities(self, dim: int) -> typing.List[AbstractCell]:
         """Get the sub-entities of the given dimension."""
-        try:
-            return [Cell(t, self._gdim) for t in self._sub_entity_types[dim]]
-        except IndexError:
-            return 0
+        return self._sub_entities
 
     def sub_entity_types(self, dim: int) -> typing.List[AbstractCell]:
         """Get the unique sub-entity types of the given dimension."""
-        try:
-            return [Cell(t, self._gdim) for t in set(self._sub_entity_types[dim])]
-        except IndexError:
-            return 0
+        return self._sub_entity_types
 
     def _lt(self, other: Self) -> bool:
         return self._cellname < other._cellname
--- a/ufl/cell.py
+++ b/ufl/cell.py
@@ -91,10 +91,6 @@ class AbstractCell(UFLObject):
         """Get the number of faces."""
         return self.num_sub_entities(2)
 
-    def num_volumes(self) -> int:
-        """Get the number of faces."""
-        return self.num_sub_entities(3)
-
     def num_facets(self) -> int:
         """Get the number of facets.
 
@@ -106,7 +102,7 @@ class AbstractCell(UFLObject):
     def num_ridges(self) -> int:
         """Get the number of ridges.
 
-        Facets are entities of dimension tdim-2.
+        Ridges are entities of dimension tdim-2.
         """
         tdim = self.topological_dimension()
         return self.num_sub_entities(tdim - 2)
@@ -114,7 +110,7 @@ class AbstractCell(UFLObject):
     def num_peaks(self) -> int:
         """Get the number of peaks.
 
-        Facets are entities of dimension tdim-3.
+        Peaks are entities of dimension tdim-3.
         """
         tdim = self.topological_dimension()
         return self.num_sub_entities(tdim - 3)
@@ -131,10 +127,6 @@ class AbstractCell(UFLObject):
         """Get the faces."""
         return self.sub_entities(2)
 
-    def volumes(self) -> typing.List[AbstractCell]:
-        """Get the faces."""
-        return self.sub_entities(3)
-
     def facets(self) -> typing.List[AbstractCell]:
         """Get the facets.
 
@@ -146,7 +138,7 @@ class AbstractCell(UFLObject):
     def ridges(self) -> typing.List[AbstractCell]:
         """Get the ridges.
 
-        Facets are entities of dimension tdim-2.
+        Ridges are entities of dimension tdim-2.
         """
         tdim = self.topological_dimension()
         return self.sub_entities(tdim - 2)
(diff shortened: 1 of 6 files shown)


ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
 
-# Export list for ufl.classes
 __all_classes__ = ["AbstractCell", "Cell", "TensorProductCell"]
 
 
-# --- The most abstract cell class, base class for other cell types
+class AbstractCell(UFLObject):
+    """A base class for all cells."""
+    @abstractmethod
+    def topological_dimension(self) -> int:
```
garth-wells, line 32: Could we make things properties rather than methods where appropriate?

mscroggs: I was planning to do that in a separate PR.

I was aiming for this PR to need no changes (or [negligible changes](FEniCS/ffcx#547)) in FFCx, DOLFINx, TSFC, and Firedrake so we could get it merged more easily and I can start experimenting with whether its useful to implement cells in the Basix UFL wrapper.

mscroggs: I've opened #169 and #170 for things I'm planning to do following this that will change the interfaces to elsewhere

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+        if type(self) == type(other):
+            s = (self.geometric_dimension(), self.topological_dimension())
+            o = (other.geometric_dimension(), other.topological_dimension())
+            if s != o:
+                return s < o
+            return self._lt(other)
+        else:
+            if type(self).__name__ == type(other).__name__:
+                raise ValueError("Cannot order cell types with the same name")
+            return type(self).__name__ < type(other).__name__
```
wence-, line 82: Could one define a total order over cells given information about the cell complex they represent? Or is that insufficient because you might have two different cell implementations that both represent a hexahedron (say).

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
-        # Sort by gdim first, tdim next, then whatever's left
-        # depending on the subclass
-        s = (self.geometric_dimension(), self.topological_dimension())
-        o = (other.geometric_dimension(), other.topological_dimension())
-        if s != o:
-            return s < o
-        return self._ufl_hash_data_() < other._ufl_hash_data_()
+    def num_volumes(self) -> int:
+        """Get the number of faces."""
+        return self.num_sub_entities(3)
```
wence-, line 98: Subtly wrong if 4-D and higher cells are ever implemented. Should probably be `self.num_sub_entities(tdim)`.

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
-                     "tetrahedron": (4, 6, 4, 1),
-                     "prism": (6, 9, 5, 1),
-                     "pyramid": (5, 8, 5, 1),
-                     "hexahedron": (8, 12, 6, 1)}
+    def num_ridges(self) -> int:
+        """Get the number of ridges.
 
-# Mapping from cell name to topological dimension
-cellname2dim = dict((k, len(v) - 1) for k, v in num_cell_entities.items())
+        Facets are entities of dimension tdim-2.
```
wence-, line 109: docstring copy-paste?

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+        """
+        tdim = self.topological_dimension()
+        return self.num_sub_entities(tdim - 2)
 
+    def num_peaks(self) -> int:
+        """Get the number of peaks.
 
-# --- Basic cell representation classes
+        Facets are entities of dimension tdim-3.
```
wence-, line 117: "peaks" not facets?

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+from abc import abstractmethod
+
+try:
+    from typing import Self
+except ImportError:
+    # Hack to get this to work pre Python 3.11
+    # Delete this after 04 Oct 2026 (Python 3.10 end of life)
+    # After this date, can also replace "self: Self," with "self,
+    from typing import TypeVar
+    Self = TypeVar("Self", bound="AbstractCell")
```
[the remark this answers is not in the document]
mscroggs: Thanks, I didn't know about typing_extensions

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+        return self.sub_entities(2)
 
-        # The topological dimension is defined by the cell type, so
-        # the cellname must be among the known ones, so we can find
-        # the known dimension, unless we have a product cell, in which
-        # the given dimension is used
-        topological_dimension = len(num_cell_entities[cellname]) - 1
+    def volumes(self) -> typing.List[AbstractCell]:
+        """Get the faces."""
+        return self.sub_entities(3)
```
wence-, line 136: `tdim` ? Or is volume a purely 3D concept? "faces" -> "volumes"

ufl/cell.py
mscroggs: > you might have two different cell implementations that both represent a hexahedron (say).

Exactly that

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+        """Get the number of sub-entities of the given dimension."""
+        try:
+            return self._num_cell_entities[dim]
+        except IndexError:
+            return 0
+
+    def sub_entities(self, dim: int) -> typing.List[AbstractCell]:
+        """Get the sub-entities of the given dimension."""
+        try:
+            return [Cell(t, self._gdim) for t in self._sub_entity_types[dim]]
```
wence-, line 273: Could you just make these up front in the constructor rather than rebuilding every time one asks?

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+        except IndexError:
+            return 0
+
+    def sub_entity_types(self, dim: int) -> typing.List[AbstractCell]:
+        """Get the unique sub-entity types of the given dimension."""
+        try:
+            return [Cell(t, self._gdim) for t in set(self._sub_entity_types[dim])]
```
wence-, line 280: Similarly here.

ufl/cell.py
```diff
@@ -1,298 +1,456 @@
...
+    # After this date, can also replace "self: Self," with "self,
+    from typing import TypeVar
+    Self = TypeVar("Self", bound="AbstractCell")
 
-# Export list for ufl.classes
 __all_classes__ = ["AbstractCell", "Cell", "TensorProductCell"]
 
 
-# --- The most abstract cell class, base class for other cell types
+class AbstractCell(UFLObject):
```
wence-, line 27: My usual preferred approach these days for abstract interfaces is `typing.Protocol` rather than `ABC`. That said, I think because you have implementation in this class and third-party libraries will inherit from it, `Protocol` is not necessarily better.

WDYT?

wence- (commented):
Some docstring fixes required I think.

ufl/cell.py
wence-: Is it sufficient (for sorting) to only define `__lt__` or does one need to provide `__eq__` and then `@functools.total_ordering` ?

mscroggs: Good question. `__eq__` is defined as part of `UFLObject` (formerly `@attach_operators_from_hash_data`). Only this and `__lt__` were previously defined so I guess it's enough for what UFL currently does, but adding `@functools.total_ordering` probably isn't a bad idea

ufl/cell.py
```diff
@@ -1,298 +1,449 @@
...
-        if keywords and keywords != ["geometric_dimension"]:
-            raise ValueError(
-                "TensorProductCell got an unexpected keyword argument '%s'" %
-                keywords[0])
+        self._sub_entities = [[Cell(t, self._gdim) for t in se_types]
+                              for se_types in self._sub_entity_types[:-1]]
+        self._sub_entity_types = [[Cell(t, self._gdim) for t in set(se_types)]
+                                  for se_types in self._sub_entity_types[:-1]]
+        self._sub_entities.append([self])
+        self._sub_entity_types.append([self])
```
wence-, line 233: Could avoid ref-cycles by using a `weakref.proxy` here if you want.

ufl/cell.py
```diff
@@ -1,298 +1,449 @@
...
+            return 0
+
+    def sub_entities(self, dim: int) -> typing.List[AbstractCell]:
+        """Get the sub-entities of the given dimension."""
+        try:
+            return self._sub_entities[dim]
+        except IndexError:
+            return []
+
+    def sub_entity_types(self, dim: int) -> typing.List[AbstractCell]:
```
wence-, line 272: FWIW, I kind of prefer tuples to lists for what is realistically constant data. Also, it's then hashable so you can use it in sets/dicts.

ufl/cell.py
mscroggs: In the element class, I think it has to be an ABC for now, as there's some implementation in the base class.

Long term, we should maybe aim for `Protocol`s, but for now I think it's confusing to have a mixture of the two, so I'm leaning towards making them `ABC`s for now

garth-wells:
@mscroggs is this almost ready?

mscroggs:
> @mscroggs is this almost ready?

Yes, I think this one is ready

garth-wells (approved):

PR #175, opened 2023-06-29
Remove `type_extensions` dependency

### LINKED ISSUE #174: `typing_extensions` breaks Spack build
See https://github.com/FEniCS/dolfinx/actions/runs/5407842569/jobs/9826291336.

Is it essential?

garth-wells:
It introduced an extra dependency (for Python < 3.11) but used in very few places. The extra, somewhat esoteric, dependency complicates Spack builds.

In the mid-term, we can require python 3.11 rather than require type_extensions.

Fixes #174.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,6 @@ setup_requires =
     wheel
 install_requires =
     numpy
-    typing_extensions; python_version < "3.11"
 
 [options.extras_require]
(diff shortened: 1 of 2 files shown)


wence- (approved):

mscroggs:
> In the mid-term, we can require python 3.11 rather than require type_extensions.

I've opened #176 to remind us of this

PR #178, opened 2023-06-29
Add Counted mixin class and refactor form signature computation

connorjward:
In Firedrake we have been running into [form signature stability problems](https://github.com/firedrakeproject/firedrake/issues/2999) with our `Constant`s since they are implemented differently to how they are done in UFL. Specifically we need to add them to `renumbering` when the form signature is calculated.

I have refactored the `counted_init` part of UFL and replaced it with a `Counted` mixin class. This means that we can compute the renumbering for all instances of `Counted`, rather than specifically for `Coefficient`s, `Constant`s etc.


dham (approved):

connorjward:
@mscroggs please could you take a look at this? Tests appear to be failing because of some unrelated linting issues. I suspect flake8 have released a new (stricter) version.

mscroggs (approved):

PR #184, opened 2023-07-25
Implements the perp operator as a class

tommbendall:
The `perp` operation is currently implemented as function:
```
lambda u: as_vector([-u[1], u[0]])
```

This create problems for the form manipulation language used in Firedrake's Gusto package. When we try to use UFL's replacement routines on a perp-ed vector, the perp-ed vector is not recognised as an operation on the original vector.

This PR addresses this by implementing `perp` as a class `Perp`, just as other operations such as the cross product are implemented.


dham (dismissed):
Please also test for Firedrake compatibility.

mscroggs (approved):

PR #198, opened 2023-09-12
Deprecated ways of defining coefficients removed

ignacia-fp:
Two instances where coefficients were defined for old ufl files have been removed. However, there still are methods that handle `legacy_domains`, which could also be remnants of old files. I suggest to review its use, because these might be no longer in use as well.


mscroggs:
Closing this as #201 is its replacement

PR #201, opened 2023-09-12
Deprecation messages for legacy features

ignacia-fp:
Deprecation messages have been introduced when Legacy objects are used. It seems that despite the legacy elements are deprecated, many tests and functions in ufl still use these features. In some cases `FiniteElement` objects are being passed to functions and  are transformed by `arguments.py` and `coefficients.py` to a `FunctionSpace` object. The code should be deprecated in principle by December 2023, but we must review the usage of `FiniteElement` within ufl in order to complete this.

### DIFF
--- a/ufl/argument.py
+++ b/ufl/argument.py
@@ -47,8 +47,8 @@ class BaseArgument(object):
             element = function_space
             domain = default_domain(element.cell())
             function_space = FunctionSpace(domain, element)
-            warnings.warn("The use of FiniteElement will be deprecated by December 2023. "
-                          "Please, use FunctionSpace instead", DeprecationWarning)
+            warnings.warn("The use of FiniteElement as an input to Argument will be deprecated by December 2023. "
+                          "Please, use FunctionSpace instead", FutureWarning)
         elif not isinstance(function_space, AbstractFunctionSpace):
             raise ValueError("Expecting a FunctionSpace or FiniteElement.")
 
--- a/ufl/argument.py
+++ b/ufl/argument.py
@@ -11,8 +11,11 @@ classes (functions), including TestFunction and TrialFunction."""
 # Modified by Anders Logg, 2008-2009.
 # Modified by Massimiliano Leoni, 2016.
 # Modified by Cecile Daversin-Catty, 2018.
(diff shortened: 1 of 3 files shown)


ufl/argument.py
```diff
@@ -44,6 +47,8 @@ def __init__(self, function_space, number, part=None):
             element = function_space
             domain = default_domain(element.cell())
             function_space = FunctionSpace(domain, element)
+            warnings.warn("The use of FiniteElement will be deprecated by December 2023. "
+                          "Please, use FunctionSpace instead", DeprecationWarning)
```
wence-, line 51: Minor: use `FutureWarning`, not `DeprecationWarning`.

Rationale: The python interpreter does not present `DeprecationWarning`s to users by default (but does do so with `FutureWarning`s), see https://docs.python.org/3/library/warnings.html#warning-categories for details.

e.g.

`foo.py`
```python
import warnings


class X:
    def __init__(self):
        warnings.warn("This is a deprecation warning", DeprecationWarning)
        warnings.warn("This is a future warning", FutureWarning)
```
```
$ python -c "import foo; foo.X()"
/home/wence/Documents/src/doodles/python/foo.py:7: FutureWarning: This is a future warning
```
(Note how the user doesn't see the deprecation warning).

mscroggs, line 51:
```suggestion
            warnings.warn("The use of FiniteElement as an input to Argument will be deprecated by December 2023. "
                          "Please, use FunctionSpace instead", FutureWarning)
```

ufl/coefficient.py
```diff
@@ -47,6 +50,8 @@ def __init__(self, function_space, count=None):
             element = function_space
             domain = default_domain(element.cell())
             function_space = FunctionSpace(domain, element)
+            warnings.warn("The use of FiniteElement will be deprecated by December 2023. "
+                          "Please, use FunctionSpace instead", DeprecationWarning)
```
mscroggs, line 54:
```suggestion
            warnings.warn("The use of FiniteElement as an input to Coefficient will be deprecated by December 2023. "
                          "Please, use FunctionSpace instead", FutureWarning)
```

mscroggs (approved):

PR #197, opened 2023-09-12
AbstractFiniteElement

mscroggs:
- Refactored the [`FiniteElementBase`](https://github.com/FEniCS/ufl/blob/mscroggs/remove-elements/ufl/finiteelement.py) class, and renamed it to `AbstractFiniteElement` (see #170)
- Moved the current finite element implementation to `legacy/` and added warning about deprecation if they're used. FEniCSx now functions without these files but they are used to get Firedrake/TSFC tests to pass. Medium term plan is to move these files to FIAT or FInAT.
- Added `FiniteElement`, `SymmetricElement` and `MixedElement` subclasses of `AbstractFiniteElement` to be used in testing. These are not loaded in `ufl/__init__.py` as they are not intended for use by users.

Resolves #160. Close #60. Close #82.

https://github.com/FEniCS/basix/pull/703, https://github.com/FEniCS/ffcx/pull/614, and https://github.com/FEniCS/dolfinx/pull/2784 must be merged alongside this to make FEniCSx tests pass.

https://github.com/firedrakeproject/tsfc/pull/300 and https://github.com/firedrakeproject/firedrake/pull/3122 must be merged once this is merged into the Firedrake fork of UFL.

PR #208, opened 2023-09-18
Format docstrings

### LINKED ISSUE #189: Use Google style docstrings
I suggest that we move to using Google style docstrings rather than Numpy style. The Google formatted strings are usually multiple lines shorter, and work more nicely with type hints.

Example Google docstring (proposed):

```python
def element(family: str, cell: str, degree: int) -> FiniteElement:
    """"Create a finite element.

    Args:
        family: The element family
        cell: The cell
        degree: The polynomial degree
    """
```

Example Numpy docstring:

```python
def element(family: str, cell: str, degree: int) -> FiniteElement:
    """"Create a finite element.

    Parameters
    -----------
    family: str
        The element family
    cell: str
        The cell
    degree: int
        The polynomial degree
    """
```

### LINKED ISSUE #202: Enable pydocstyle checks

mscroggs:
Remove pydocstyle test excludes from setup.cfg, and set convention to google. Make all docstrings satisfy these checks.

For files in the `finiteelement/` folder, the garbage docstrings have been added to make the checks pass. This is because in another PR (#197) I'm working on removing these files, so it seems like a waste of time to write nicely formatted docstrings in these files.

Close #202. Close #189.


ufl/algorithms/analysis.py
```diff
@@ -46,8 +46,13 @@ def unique_tuple(objects):
 
 def extract_type(a, ufl_types):
     """Build a set of all objects found in a whose class is in ufl_types.
-    The argument a can be a BaseForm, Integral or Expr."""
 
+    Args:
+        a: A BaseForm, Integral or Expr
+        ufl_types: A list of UFL types
+
+    Returns: All objects found in a whose class is in ufl_type
```
garth-wells, line 54: Should this be 
```
Returns: 
    All objects found in a whose class is in ufl_type
```
in the Google docstyle, or are both ok?

PR #213, opened 2023-09-20
Remove IrreducibleInt

mscroggs:
Remove the `IrreducibleInt` hack.

It seems locally that all the tests pass without it...

This is coupled with https://github.com/FEniCS/basix/pull/702

### DIFF
--- a/.github/workflows/fenicsx-tests.yml
+++ b/.github/workflows/fenicsx-tests.yml
@@ -34,7 +34,7 @@ jobs:
 
       - name: Install Basix
         run: |
-          python3 -m pip install git+https://github.com/FEniCS/basix.git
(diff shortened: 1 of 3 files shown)


mscroggs:
> It's a long time, but I think this hack was put in because the simplex-based degree estimation code under-estimates the degrees of integrals on hypercubes (for reasons that I do not recall 😞 )

Yes, the docstring of `_reduce_degree` says that's done but also not why. I think I've achieved the same result by checking the cell type inside `_reduce_degree`:

https://github.com/FEniCS/ufl/blob/9e5884c2b767cf7d6325120d9843f5df76e2ac7f/ufl/algorithms/estimate_degrees.py#L99

garth-wells (approved):

PR #214, opened 2023-09-21
Update action differentiation

nbouziani:
Prevent action differentiation, i.e. when we have `derivative(Action(..., ...), ...)`, to eagerly expand derivatives.


test/test_duals.py
```diff
@@ -274,11 +274,12 @@ def test_differentiation():
...
     Ac = Action(w, u)
     dAcdu = derivative(Ac, u)
-    assert dAcdu == action(adjoint(derivative(w, u)), u) + action(w, derivative(u, u))
+    assert dAcdu == (action(adjoint(derivative(w, u), derivatives_expanded=True), u, derivatives_expanded=True)
+                     + action(w, derivative(u, u), derivatives_expanded=True))
 
     dAcdu = expand_derivatives(dAcdu)
     # Since dw/du = 0
-    assert dAcdu == Action(w, v)
+    assert dAcdu == 1 * Action(w, v)
```
mscroggs, line 282: Is `1 * ` really needed here?

nbouziani: It's for type compatibility, i.e. to get a `FormSum` object. I have updated the code so that those simple cases get simplified.

PR #220, opened 2023-10-15
Remove deprecated functionality (attach_operators_from_hash_data).

jorgensd:
The deprecation was introduced 4 months ago (https://github.com/FEniCS/ufl/pull/168). Since all the core objects of UFL is using it (Mesh and FunctionSpace) it means that pytest is throwing tons of deprecation warnings at import:
```python
python3 -W error::DeprecationWarning -c "import ufl.Mesh"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/root/shared/ufl/__init__.py", line 265, in <module>
    from ufl.domain import as_domain, AbstractDomain, Mesh, MeshView, TensorProductMesh
  File "/root/shared/ufl/domain.py", line 190, in <module>
    class TensorProductMesh(AbstractDomain):
  File "/root/shared/ufl/core/ufl_type.py", line 56, in attach_operators_from_hash_data
    warnings.warn("attach_operators_from_hash_data deprecated, please use UFLObject instead.", DeprecationWarning)
DeprecationWarning: attach_operators_from_hash_data deprecated, please use UFLObject instead.
```


jorgensd:
@mscroggs  seems like adding `__str__` as an Abstract class might not be a great idea, as every object would have to overload it.

PR #227, opened 2023-10-23
We do require pip>=22.3 for editable installs (PEP 660).

jorgensd:
This does not quite resolve issues with users trying to install with an outdated pip version, as the upgrade happens at runtime, and doesn't change the current version of pip that is executed. PEP 660: https://peps.python.org/pep-0660/


jhale:
This is the best we can do I think.

garth-wells:
The shouldn't really be done - this version of pip isn't needed for standard installs. Unnecessary dependency pinning complicates packaging.

Users who need something niche (like editable installs) can easy take care of the pip version themselves.

PR #236, opened 2023-11-02
Dolci/integral symbol

Ig-dolci:
In a recent change to UFL, `subdomain_id` was transformed from a single value to a tuple. This broke some assumptions in ufl2unicode that I have fixed here.


ksagiyam:
@Ig-dolci This needs to be rebased on FEniCS:main; otherwise we see many "merge_upstream" commits.

PR #242, opened 2023-11-16
`FormSum.coefficients()` should be ordered deterministically

JHopeCollins:
`Form.coefficients()` and `FormSum.coefficients()` both use `set` to produce a unique collection of the coefficients. However, `set` does not have deterministic ordering. `Form._analyze_form_arguments()` uses `sorted` to make sure that the ordering is deterministic across MPI ranks and across multiple programme executions, but `FormSum._analyze_form_arguments()` doesn't.

This means that the ordering is sometimes different between different MPI ranks, which causes parallel hangs in MG code when a `Function` is sent to `inject` on one rank and a `Cofunction` is sent to `restrict` on another rank.

This PR just adds orders `FormSum.coefficients()` in the same way as `Form.coefficients()`


ufl/form.py
```diff
@@ -784,8 +784,10 @@ def _analyze_form_arguments(self):
         for component in self._components:
             arguments.extend(component.arguments())
             coefficients.extend(component.coefficients())
-        self._arguments = tuple(set(arguments))
-        self._coefficients = tuple(set(coefficients))
+        self._arguments = tuple(
+            sorted(set(arguments), key=lambda x: x.number()))
+        self._coefficients = tuple(
+            sorted(set(coefficients), key=lambda x: x.count()))
```
wence-, line 790: Can `Coefficients` and `CoFunctions` end up with the same `count`?

JHopeCollins:
Closed here and reopened on the Firedrake/ufl
[https://github.com/firedrakeproject/ufl/pull/47](https://github.com/firedrakeproject/ufl/pull/47)

PR #243, opened 2023-11-17
`FormSum.coefficients` should be ordered deterministically the same as `Form.coefficients`

JHopeCollins:
`Form.coefficients()` and `FormSum.coefficients()` both use `set` to produce a unique collection of the coefficients. However, `set` does not have deterministic ordering. `Form._analyze_form_arguments()` uses `sorted` to make sure that the ordering is deterministic across MPI ranks and across multiple programme executions, but `FormSum._analyze_form_arguments()` doesn't.

This means that the ordering is sometimes different between different MPI ranks, which causes parallel hangs in MG code when a `Function` is sent to `inject` on one rank and a `Cofunction` is sent to `restrict` on another rank.

This PR just adds orders `FormSum.coefficients()` in the same way as `Form.coefficients()`

### DIFF
--- a/ufl/form.py
+++ b/ufl/form.py
@@ -784,8 +784,11 @@ class FormSum(BaseForm):
         for component in self._components:
             arguments.extend(component.arguments())
             coefficients.extend(component.coefficients())
-        self._arguments = tuple(set(arguments))
-        self._coefficients = tuple(set(coefficients))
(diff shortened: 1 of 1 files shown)


nbouziani (approved):
This makes sense to me as long as it passes tests.

dham (approved):

PR #246, opened 2023-11-23
Expunge ufl_domain

pbrubeck:


mscroggs:
Can we add a deprecation warning to the ufl_domain function, then we can remove it in a couple of months' time?

jorgensd:
> Can we add a deprecation warning to the ufl_domain function, then we can remove it in a couple of months' time?

We have this in certain places already:
```bash
   /usr/local/lib/python3.10/dist-packages/ufl/core/expr.py:269: DeprecationWarning: Expr.ufl_domain() is deprecated, please use extract_unique_domain(expr) instead.
    warnings.warn("Expr.ufl_domain() is deprecated, please "
```
so I think that is a sensible action.

dham:
> > Can we add a deprecation warning to the ufl_domain function, then we can remove it in a couple of months' time?
> 
> We have this in certain places already:
> 
> ```shell
>    /usr/local/lib/python3.10/dist-packages/ufl/core/expr.py:269: DeprecationWarning: Expr.ufl_domain() is deprecated, please use extract_unique_domain(expr) instead.
>     warnings.warn("Expr.ufl_domain() is deprecated, please "
> ```
> 
> so I think that is a sensible action.

I agree that this is sensible, however our experience is that users ignore warnings. I'd therefore suggest that we then progress to making `url_domain` raise a hard error for a period before we actually delete it. That way users will get documentation on how to fix their code rather than just a missing function.

dham (approved):

PR #252, opened 2024-01-30
Emmarothwell1/restricted function space

dham:
Add support for function space labels in order to support symbolically different function spaces on the same element and domain. An example are restricted function spaces.


ufl/functionspace.py
```diff
@@ -50,11 +50,16 @@ def __init__(self, domain, element):
             else:
                 if element.cell != domain_cell:
                     raise ValueError("Non-matching cell of finite element and domain.")
-
+
```
dham, line 53:
```suggestion

```

ufl/functionspace.py
```diff
@@ -80,6 +85,7 @@ def _ufl_hash_data_(self, name=None):
         name = name or "BaseFunctionSpace"
         domain = self.ufl_domain()
         element = self.ufl_element()
+        bdata = self.label()
```
dham, line 88: I guess we don't need this variable, just directly use `self.label` below.

ufl/functionspace.py
```diff
@@ -88,13 +94,14 @@ def _ufl_hash_data_(self, name=None):
...
             edata = element._ufl_hash_data_()
-        return (name, ddata, edata)
+        return (name, ddata, edata, bdata)
 
     def _ufl_signature_data_(self, renumbering, name=None):
         """UFL signature data."""
         name = name or "BaseFunctionSpace"
         domain = self.ufl_domain()
         element = self.ufl_element()
+        bdata = self.label()
```
dham, line 104: as above.

ufl/functionspace.py
```diff
@@ -50,11 +50,15 @@ def __init__(self, domain, element):
...
                 if element.cell != domain_cell:
                     raise ValueError("Non-matching cell of finite element and domain.")
-
         AbstractFunctionSpace.__init__(self)
+        self._label = label
         self._ufl_domain = domain
         self._ufl_element = element
 
+    def label(self):
+        """Return label of boundary domains to differentiate restricted and unrestricted"""
```
mscroggs, line 59:
```suggestion
        """Return label of boundary domains to differentiate restricted and unrestricted."""
```

PR #259, opened 2024-02-19
Remove print statement

jorgensd:


jorgensd:
Close in favor of #258

PR #271, opened 2024-04-19
Update apply_restrictions

mscroggs:
Replace multifunction for apply_restrictions with a class method (defined in a protocol). Caching of results is now achieved using `functools.lru_cache` rather than inside the multifunction.

This will allow `apply_restrictions` to be implemented for types defined outside UFL without having to add the implementation of this to `algorithms/apply_restrictions.py`.

As far as users are concerned, there should be no noticable change.

I propose that we aim to make this change for all multifunctions (https://github.com/FEniCS/ufl/issues/273).

### DIFF
--- a/ufl/algorithms/apply_restrictions.py
+++ b/ufl/algorithms/apply_restrictions.py
@@ -10,11 +10,11 @@ restrictions in a form towards the terminals.
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 import typing
-import warnings
 from abc import abstractmethod
 from typing import Protocol
 
 from ufl.algorithms.map_integrands import map_integrand_dags
+from ufl.core import caching
 from ufl.corealg.multifunction import MultiFunction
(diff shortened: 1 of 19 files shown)


ufl/algorithms/apply_restrictions.py
```diff
@@ -9,178 +9,39 @@
...
-    integral_types = [
-        k for k in integral_type_to_measure_name.keys() if k.startswith("interior_facet")
-    ]
-    rules = RestrictionPropagator()
-    return map_integrand_dags(rules, expression, only_integral_type=integral_types)
+    warnings.warn(
+        "apply_restrictions(expression) is deprecated and will be removed "
+        "after December 2024. Please, use expression.apply_restrictions() instead",
+        FutureWarning,
+    )
```
mscroggs, line 43: Do we want to deprecate this function and encourage use of `expression.apply_restrictions()` instead? Or do we want to keep it?

PR #278, opened 2024-04-25
Remove unnecessary pip pinning

garth-wells:
See https://github.com/FEniCS/ufl/pull/227#issuecomment-2076754513.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=62", "wheel", "pip>=22.3"]
+requires = ["setuptools>=62", "wheel"]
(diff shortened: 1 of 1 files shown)


jorgensd:
Why do we want to let users use old versions of pip? Is there some specific systems that haven’t updated their pip version for 2 years?

garth-wells:
> Why do we want to let users use old versions of pip? Is there some specific systems that haven’t updated their pip version for 2 years?

Why should we care? 

We should not pin versions unnecessarily because it complicates packaging.

jhale:
I don't even think pip is necessary here - it should be possible to install all of our packages using any PEP-compliant tool, e.g. `poetry`.

PR #295, opened 2024-06-20
Use subdegree instead of superdegree to check cell bendy-ness

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/3612

To quote @dham:

> The reason subdegree is right is that you care about whether the edges are straight, not whether there are any quadratic functions on the interior.

### DIFF
--- a/ufl/algorithms/apply_geometry_lowering.py
+++ b/ufl/algorithms/apply_geometry_lowering.py
@@ -304,7 +304,7 @@ class GeometryLoweringApplier(MultiFunction):
 
         domain = extract_unique_domain(o)
 
-        if domain.ufl_coordinate_element().embedded_superdegree > 1:
+        if domain.ufl_coordinate_element().embedded_subdegree > 1:
             # Don't lower bendy cells, instead leave it to form compiler
             warnings.warn("Only know how to compute cell edge lengths of P1 or Q1 cell.")
             return o
@@ -329,7 +329,7 @@ class GeometryLoweringApplier(MultiFunction):
 
         domain = extract_unique_domain(o)
 
-        if domain.ufl_coordinate_element().embedded_superdegree > 1:
+        if domain.ufl_coordinate_element().embedded_subdegree > 1:
             # Don't lower bendy cells, instead leave it to form compiler
             warnings.warn("Only know how to compute cell diameter of P1 or Q1 cell.")
             return o
@@ -366,7 +366,7 @@ class GeometryLoweringApplier(MultiFunction):
         if domain.ufl_cell().topological_dimension() < 3:
             raise ValueError("Facet edge lengths only make sense for topological dimension >= 3.")
 
-        elif domain.ufl_coordinate_element().embedded_superdegree > 1:
+        elif domain.ufl_coordinate_element().embedded_subdegree > 1:
             # Don't lower bendy cells, instead leave it to form compiler
             warnings.warn("Only know how to compute facet edge lengths of P1 or Q1 cell.")
             return o


mscroggs:
I'm surprised this makes any difference, as for P and Q spaces, the embedded sub- and superdegrees are the same.

At least, they're the same for the elements coming from Basix. Maybe this isn't true for elements from FInAT?

connorjward:
> I'm surprised this makes any difference, as for P and Q spaces, the embedded sub- and superdegrees are the same.
> 
> At least, they're the same for the elements coming from Basix. Maybe this isn't true for elements from FInAT?

They differ for tensor product elements. We encountered the issue on extruded meshes.

wence-:
> At least, they're the same for the elements coming from Basix.

They shouldn't be for `Q`, no? e.g. Q1 on quadrilaterals has a basis $\text{span} \\{1, x, y, x y \\}$ which has an embedding superdegree of 2.

mscroggs:
Embedded degrees are defined in terms of the span of Lagrange on the current cell:

https://github.com/FEniCS/ufl/blob/ea144b0bd8304776818515ec2b2e6033c63c1d87/ufl/finiteelement.py#L70-L83

This is important for pyramids, where rational functions like $\frac{x}{1-z}$ are included in the polysets, so talking about standard polynomial degrees doesn't make sense.

mscroggs:
Longer term, I think we should add information about the traces of elements (#298), as I think we should be asking here whether the traces on the edge are linear, which the embedded degrees don't really capture. But happy to merge this in the meantime

wence-:
> Embedded degrees are defined in terms of the span of Lagrange on the current cell:
> 
[...]
> 
> This is important for pyramids, where rational functions like x1−z are included in the polysets, so talking about standard polynomial degrees doesn't make sense.

I agree that if the basis is non-polynomial it doesn't make much sense. 

However, my reading of the docstring you quoted still suggests that Q1 should be advertising an `embedded_superdegree` of 2. Since the complete polynomial space that spans the Q1 basis is of degree 2. Am I misunderstanding?

mscroggs:
> However, my reading of the docstring you quoted still suggests that Q1 should be advertising an `embedded_superdegree` of 2. Since the complete polynomial space that spans the Q1 basis is of degree 2. Am I misunderstanding?

My intended reading of "This returns the degree of the lowest degree Lagrange space such  that the polynomial space of the Lagrange space is a superspace of this element's polynomial space" is "This returns the degree of the lowest degree Lagrange space **on the cell on which this element is defined** such that the polynomial space of the Lagrange space is a superspace of this element's polynomial space", ie for elements on quads, it returns the degree of the Lagrange/Q space in which it's embedded.

We were discussing this at PDESoft, and I was leaning towards your reading being better. But since thinking about pyramids I'm less sure, and remember why I made it this way.

mscroggs:
> Lagrange/Q space

Also this assumes that everyone overloads the name "Lagrange" in the same way I do which is in no way a safe assumption.

wence-:
> Also this assumes that everyone overloads the name "Lagrange" in the same way I do which is in no way a safe assumption.

Yeah, I think this is probably one aspect of the confusion. I suspect it is better to talk in terms of spans of polynomial sets, rather than names for this.

> But since thinking about pyramids I'm less sure, and remember why I made it this way.

AIUI (and I haven't really done anything with pyramids), if these have rational basis functions then there is no polynomial space that spans the basis. And so one should return `None`.

So I'm not quite sure how the "complete polynomial space" reading conflicts: there is no such space, so one returns None.

mscroggs:
> So I'm not quite sure how the "complete polynomial space" reading conflicts: there is no such space, so one returns None.

It can still be useful to know which "degree" the elements on pyramids are included in, for example for telling if the edges of the pyramid are straight.

Thought to return to later: If #298 is done, do we still need embedded degrees?

PR #303, opened 2024-08-21
introduce MixedMesh class

ksagiyam:
Introduce `MixedMesh` class for multi-mesh problems.

Note that `extract_domains()` now uses more robust `sort_domains()` inside, so it might behave slightly differently.


ufl/domain.py
```diff
@@ -126,6 +145,69 @@ def _ufl_sort_key_(self):
...
 
+    @property
+    def meshes(self):
+        """Return the component meshes."""
+        return (self,)
+
+
+@attach_ufl_id
+class MixedMesh(AbstractDomain, UFLObject):
+    """Symbolic representation of a mixed mesh."""
```
ksagiyam, line 156: doctoring

ksagiyam:
@jpdean @jorgensd This is what we briefly discussed at PDESoft. Could you have a look?

ufl/algorithms/analysis.py
```diff
@@ -198,19 +200,22 @@ def extract_base_form_operators(a):
...
+    separately for large forms, and has more validation built in.
 
     Args:
         a: A BaseForm, Integral or Expr
     """
     # Extract lists of all BaseArgument and BaseCoefficient instances
-    base_coeff_and_args = extract_type(a, (BaseArgument, BaseCoefficient))
-    arguments = [f for f in base_coeff_and_args if isinstance(f, BaseArgument)]
-    coefficients = [f for f in base_coeff_and_args if isinstance(f, BaseCoefficient)]
+    base_coeff_and_args_and_gq = extract_type(a, (BaseArgument, BaseCoefficient, GeometricQuantity))
```
jorgensd, line 213: Could we just call this list `base_types`, as it is quite explicit from the function call what it does, and it makes the latter code more readable.

ufl/algorithms/analysis.py
```diff
@@ -226,20 +231,36 @@ def extract_arguments_and_coefficients(a):
...
             "Found different coefficients with same counts.\n"
-            "The arguments found are:\n" + "\n".join(f"  {c}" for c in coefficients)
+            "The Coefficients found are:\n" + "\n".join(f"  {c}" for c in coefficients)
+        )
+
+    # Build count: instance mappings, should be one to one
+    gqcounts = {}
+    for gq in geometric_quantities:
+        if not isinstance(gq._domain, Mesh):
+            raise TypeError(f"{gq}._domain must be a Mesh: got {gq._domain}")
```
jorgensd, line 241: What else com `gq._domain` be? is this relying on the deprecated definition that a `domain` can be defined just through a `ufl.Cell`?

test/test_mixed_function_space_with_mixed_mesh.py
```diff
@@ -0,0 +1,81 @@
...
+    cell = triangle
+    elem0 = FiniteElement("Lagrange", cell, 1, (), identity_pullback, H1)
+    elem1 = FiniteElement("Brezzi-Douglas-Marini", cell, 1, (2,), contravariant_piola, HDiv)
+    elem2 = FiniteElement("Discontinuous Lagrange", cell, 0, (), identity_pullback, L2)
+    elem = MixedElement([elem0, elem1, elem2])
+    mesh0 = Mesh(FiniteElement("Lagrange", cell, 1, (2,), identity_pullback, H1), ufl_id=100)
+    mesh1 = Mesh(FiniteElement("Lagrange", cell, 1, (2,), identity_pullback, H1), ufl_id=101)
+    mesh2 = Mesh(FiniteElement("Lagrange", cell, 1, (2,), identity_pullback, H1), ufl_id=102)
+    domain = MixedMesh(mesh0, mesh1, mesh2)
+    V = FunctionSpace(domain, elem)
```
jorgensd, line 35: So we now have two ways of creating a `FunctionSpace` with `MixedElement`s.
1. A single mesh of class `Mesh` with a `MixedElement`, where `MixedElement` can have arbitrary many elements
2. A `MixedMesh` consisting of `M` meshes, and a `MixedElement` consisting of `M` elements.

Do we need additional checks in the `FunctionSpace` constructor to ensure that this is satisfied (instead of catching this at a later instance, inside for instance `GenericDerivativeRuleset.reference_value`.

I would also like to highlight that the API for `MixedElement` and `MixedMeshes` differs, as one passes a list, while the other unrolls a list, I think we should be consistent, and use lists in both places

ufl/algorithms/analysis.py
ksagiyam: Renamed it `terminals`.

ufl/algorithms/analysis.py
ksagiyam: No. It checks if `gq._domain` is `Mesh` and not `MixedMesh`.

test/test_mixed_function_space_with_mixed_mesh.py
ksagiyam: Added checks in `FunctionSpace` constructor and changed `MixedMesh` API as suggested.  Thanks.

ufl/algorithms/apply_derivatives.py
```diff
@@ -655,16 +668,52 @@ def reference_value(self, o):
...
-        K = JacobianInverse(domain)
-        Do = grad_to_reference_grad(o, K)
-        return Do
+        domain = extract_unique_domain(f, expand_mixed_mesh=False)
+        if isinstance(domain, MixedMesh):
+            element = f.ufl_function_space().ufl_element()
+            if element.num_sub_elements != len(domain):
+                raise RuntimeError(f"{element.num_sub_elements} != {len(domain)}")
+            g = ReferenceGrad(o)
+            vsh = g.ufl_shape[:-1]
```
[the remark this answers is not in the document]
ksagiyam: Turned out `vsh` could be removed, so I did.

You are right. Currently, `MixedMesh` can only be composed of meshes of the same cell_type, so topological dimension is well defined. I edited the doc string of `MixedMesh` to make this point clear.

ksagiyam:
Updated PR summary.

Added more checks.

Fixed `MixedPullback.physical_value_shape()`.

ksagiyam:
Can I have another round of review on this?

ufl/algorithms/analysis.py
```diff
@@ -198,19 +200,20 @@ def extract_base_form_operators(a):
...
+    """Build three sorted lists of all arguments, coefficients, and geometric quantities in `a`.
 
-    This function is faster than extract_arguments + extract_coefficients
-    for large forms, and has more validation built in.
+    This function is faster than extracting each type of terminal
+    separately for large forms, and has more validation built in.
 
     Args:
         a: A BaseForm, Integral or Expr
     """
```
jorgensd, line 211: Could we add typing of input/output here?

ufl/algorithms/apply_derivatives.py
```diff
@@ -84,6 +86,17 @@
...
 
+def flatten_domain_element(domain, element):
+    """Return the flattened (domain, element) pairs for mixed domain problems."""
+    if not isinstance(domain, MixedMesh):
+        return ((domain, element),)
+    flattened = ()
+    assert len(domain) == len(element.sub_elements)
+    for d, e in zip(domain, element.sub_elements):
+        flattened += flatten_domain_element(d, e)
+    return flattened
```
jorgensd, line 97: Typing + specification of what is returned if the problem is not a "mixed domain problem"/

ufl/algorithms/apply_derivatives.py
```diff
@@ -657,16 +670,51 @@ def reference_value(self, o):
...
+                    if gdim != self._var_shape[0]:
+                        raise RuntimeError(f"{gdim} != {self._var_shape[0]}")
+                    for idx in range(int(np.prod(esh))):
+                        for i in range(gdim):
+                            temp = Zero()
+                            for j in range(rdim):
+                                temp += g[(dofoffset + idx,) + (j,)] * K[j, i]
+                            components.append(temp)
+                dofoffset += int(np.prod(esh))
+            return as_tensor(np.asarray(components).reshape(g.ufl_shape[:-1] + self._var_shape))
```
jorgensd, line 710: Does one need the conditional when `flatten_domain_element` works for non-mixed meshes.

ufl/algorithms/apply_derivatives.py
```diff
@@ -657,16 +670,51 @@ def reference_value(self, o):
...
+                        raise RuntimeError(f"{rdim} != {ref_dim}")
+                    if gdim != self._var_shape[0]:
+                        raise RuntimeError(f"{gdim} != {self._var_shape[0]}")
+                    for idx in range(int(np.prod(esh))):
+                        for i in range(gdim):
+                            temp = Zero()
+                            for j in range(rdim):
+                                temp += g[(dofoffset + idx,) + (j,)] * K[j, i]
+                            components.append(temp)
+                dofoffset += int(np.prod(esh))
```
jorgensd, line 709: Some documentation of the logic within this loop would be nice.
I guess the idea is:
- For a mixed domain problem, accumulate all derivatives in expanded index notation.
Please correct me if I am mistaken.

I think one thing that would help is to define `np.prod(esh)` as a variable, as it is used in multiple loops and to accumulate the counter.

Also, I think you do not need the
```
if esh == ():
    esh = (1,)
```
as `int(np.prod(()))` ref: https://numpy.org/doc/2.1/reference/generated/numpy.prod.html
> The product of an empty array is the neutral element 1

One could add a guard against future changing behavior by adding
```python
esh = e.reference_value_shape
num_components = int(np.prod(esh))
assert num_components > 0
```

ufl/algorithms/apply_derivatives.py
jorgensd, line 710: Is using `numpy.asarray` safe here?

ufl/algorithms/apply_derivatives.py
```diff
@@ -678,10 +726,43 @@ def reference_grad(self, o):
...
+        domain = extract_unique_domain(f, expand_mixed_mesh=False)
+        if isinstance(domain, MixedMesh):
+            if not f._ufl_is_in_reference_frame_:
+                raise RuntimeError("Expecting a reference frame type")
+            while not f._ufl_is_terminal_:
+                (f,) = f.ufl_operands
+            element = f.ufl_function_space().ufl_element()
+            if element.num_sub_elements != len(domain):
+                raise RuntimeError(f"{element.num_sub_elements} != {len(domain)}")
+            g = ReferenceGrad(o)
```
jorgensd, line 738: Same comment as for previous function

ufl/algorithms/compute_form_data.py
```diff
@@ -262,6 +262,14 @@ def compute_form_data(
 
     The default arguments configured to behave the way old FFC expects.
     """
+    # Currently, only integral_type="cell" can be used with MixedMesh.
+    for integral in form.integrals():
+        if integral.integral_type() != "cell":
+            all_domains = extract_domains(integral.integrand(), expand_mixed_mesh=False)
+            if any(isinstance(m, MixedMesh) for m in all_domains):
+                raise NotImplementedError("""
+                    Only integral_type="cell" can be used with MixedMesh""")
```
jorgensd, line 271: Maybe add what cell type you got as input to error message.

ufl/domain.py
```diff
@@ -50,6 +50,33 @@ def topological_dimension(self):
...
+        """Return iterable component meshes."""
+        return iter(self.meshes)
+
+    def iterable_like(self, element):
+        """Return iterable object that is iterable like ``element``."""
+        raise NotImplementedError("iterable_like() method not implemented")
+
+    def can_make_function_space(self, element):
+        """Check whether this mesh can make a function space with ``element``."""
+        raise NotImplementedError("check_compatibility() method not implemented")
```
jorgensd, line 78:
```suggestion
        raise NotImplementedError("can_make_function_space() method not implemented")
```

ufl/domain.py
```diff
@@ -126,6 +153,119 @@ def _ufl_sort_key_(self):
         typespecific = (self._ufl_id, self._ufl_coordinate_element)
         return (self.geometric_dimension(), self.topological_dimension(), "Mesh", typespecific)
 
+    @property
+    def meshes(self):
+        """Return the component meshes."""
+        return (self,)
+
+    def iterable_like(self, element):
```
jorgensd, line 161: Typing-hinting for new functions in UFL would be nice.

jorgensd: This should be type-hinted as it has assumptions on the fact that element has the property `sub_elements`.

ufl/domain.py
```diff
@@ -126,6 +153,119 @@ def _ufl_sort_key_(self):
...
+        typespecific = (self._ufl_id,)
+        return (self.geometric_dimension(), self.topological_dimension(), "MixedMesh", typespecific)
+
+    @property
+    def meshes(self):
+        """Return the component meshes."""
+        return self._meshes
+
+    def iterable_like(self, element):
+        """Return iterable object that is iterable like ``element``."""
```
jorgensd, line 256: Assumes that you have `num_sub_elements`, so should be type-hinted.

ufl/domain.py
jorgensd: Why not use `num_sub_elments` as in `mixedmesh`?

ufl/domain.py
```diff
@@ -126,6 +153,119 @@ def _ufl_sort_key_(self):
...
+        if len(self) != element.num_sub_elements:
+            raise RuntimeError(f"""len(self) ({len(self)}) !=
+                element.num_sub_elements ({element.num_sub_elements})""")
+        return self
+
+    def can_make_function_space(self, element):
```
jorgensd, line 262: Type-hinting.

ufl/domain.py
```diff
@@ -190,26 +330,34 @@ def as_domain(domain):
...
+        return domain
 
 
 def sort_domains(domains):
     """Sort domains in a canonical ordering."""
     return tuple(sorted(domains, key=lambda domain: domain._ufl_sort_key_()))
 
 
-def join_domains(domains):
+def join_domains(domains, expand_mixed_mesh=True):
```
jorgensd, line 348: Add comment about what expand mixed mesh true/false means + type-hinting.

ufl/domain.py
```diff
@@ -226,20 +374,26 @@ def join_domains(domains):
 # TODO: Move these to an analysis module?
 
 
-def extract_domains(expr):
+def extract_domains(expr, expand_mixed_mesh=True):
```
jorgensd, line 377: Add documentation regarding keyword argument `expand_mixed_mesh`.

ufl/domain.py
```diff
@@ -226,20 +374,26 @@ def join_domains(domains):
...
+        for t in traverse_unique_terminals(expr):
+            domainlist.extend(t.ufl_domains())
+        return sort_domains(join_domains(domainlist, expand_mixed_mesh=expand_mixed_mesh))
 
 
-def extract_unique_domain(expr):
+def extract_unique_domain(expr, expand_mixed_mesh=True):
     """Return the single unique domain expression is defined on or throw an error."""
-    domains = extract_domains(expr)
+    domains = extract_domains(expr, expand_mixed_mesh=expand_mixed_mesh)
```
jorgensd, line 396: Add kwarg docstring

jhale:
Copied from Slack:

The implementation looks OK to me (quick look), but the tests look a bit light. In terms of the naming, I was also confused by the use of Mixed. At least when we’re talking about MixedElement, aside from it having historical meaning, my impression was that we are talking about using the Cartesian product on the elements of the list and maintaining ordering. My understanding of MixedMesh is that we are maintaining the same ordering, but I didn’t get the impression it was a Cartesian product - consequently, the used of Mixed wasn’t clear to me. This is aside from the idea that Mixed Mesh has no traditional meaning in the literature - when the idea was first proposed I thought it was related to having a mesh with mixed cell type, to illustrate the confusion.

Of course these are ‘misconceptions’ on my part, but I’m just trying to get across the confusion that appeared in my own mind, so we can discuss if different naming of the same concept might be appropriate.

connorjward (commented):
I am happy to be argued down but personally I think the abstraction is a bit strange. Why can we not have, instead of
```py
    cell = triangle
    elem0 = FiniteElement("Lagrange", cell, ...)
    elem1 = FiniteElement("Brezzi-Douglas-Marini", cell, ...)
    elem2 = FiniteElement("Discontinuous Lagrange", cell, ...)
    elem = MixedElement([elem0, elem1, elem2])
    mesh0 = Mesh(FiniteElement("Lagrange", cell, ...))
    mesh1 = Mesh(FiniteElement("Lagrange", cell, ...))
    mesh2 = Mesh(FiniteElement("Lagrange", cell, ...))
    domain = MixedMesh([mesh0, mesh1, mesh2])
    V = FunctionSpace(domain, elem)
```
something like
```py
    cell = triangle
    elem0 = FiniteElement("Lagrange", cell, ...)
    mesh0 = Mesh(FiniteElement("Lagrange", cell, ...))
    V0 = FunctionSpace(mesh0, elem0)

    elem1 = FiniteElement("Brezzi-Douglas-Marini", cell, ...)
    mesh1 = Mesh(FiniteElement("Lagrange", cell, ...))
    V1 = FunctionSpace(mesh1, elem1)

    elem2 = FiniteElement("Discontinuous Lagrange", cell, ...)
    mesh2 = Mesh(FiniteElement("Lagrange", cell, ...))
    V2 = FunctionSpace(mesh2, elem2)

    V = MixedFunctionSpace([V0, V1, V2])
```
where we have a `MixedFunctionSpace` instead of a `MixedMesh`.

More generally what is the difference between a `FunctionSpace` with a `MixedElement` and a `MixedFunctionSpace` using non-`Mixed` elements?

jorgensd:
> ```python
>     cell = triangle
>     elem0 = FiniteElement("Lagrange", cell, ...)
>     mesh0 = Mesh(FiniteElement("Lagrange", cell, ...))
>     V0 = FunctionSpace(mesh0, elem0)
> 
>     elem1 = FiniteElement("Brezzi-Douglas-Marini", cell, ...)
>     mesh1 = Mesh(FiniteElement("Lagrange", cell, ...))
>     V1 = FunctionSpace(mesh1, elem1)
> 
>     elem2 = FiniteElement("Discontinuous Lagrange", cell, ...)
>   
> ```

This is what we use in DOLFINx at the moment.

I think the current  mixed-mesh abstraction is going to be confusing when we introduce meshes with multiple cell types (ie mix of quadrilateral and triangle, or hex, yet, prism).

ksagiyam:
@jorgensd I do not think merely allowing for `Coefficient(MixedFunctionSpace)` will solve all issues; as long as `MixedFunctionSpace` does not subclass `FunctionSpace` and does not have `ufl_domain`, I think we will need to special case functions/methods in UFL for `MixedFunctionSpace` just as `Coefficients` function currently does. As I mentioned, introduction of `MixedMesh` allows UFL to be unaware of the mixed spaces in the most part, and it only needs to special case methods at the very bottom, i.e. `GradRuleset.reference_value()`, `GradRuleset.reference_grad()`, and `pullback.apply()`.

I did not mean nested mixed spaces by "multiple meshes + multiple spaces". I just meant (meshA, elemA) x (meshB, elemB). Though we do not use nested mixed spaces in Firedrake, some functions such as `flatten_domain_element` are written with this in mind as you might have noticed. I do not think consideration of such nested mixed spaces would affect the mixed-problem design in UFL, though.

@jhale I think the use of `Mixed` in `MixedMesh` is consistent with the use of `Mixed` in `MixedElement`. If we consider assembling (i, j)-block of a matrix, we can view it as mixing `mixed_mesh[i]` and `mixed_mesh[j]` as well as mixing `mixed_element[i]` and `mixed_element[j]`. 

But I agree that it is confusing. As @dham suggested, `MeshSequence` or `MultiMesh` is also fine by me.

jorgensd:
> @jorgensd I do not think merely allowing for `Coefficient(MixedFunctionSpace)` will solve all issues; as long as `MixedFunctionSpace` does not subclass `FunctionSpace` and does not have `ufl_domain`, I think we will need to special case functions/methods in UFL for `MixedFunctionSpace` just as `Coefficients` function currently does. As I mentioned, introduction of `MixedMesh` allows UFL to be unaware of the mixed spaces in the most part, and it only needs to special case methods at the very bottom, i.e. `GradRuleset.reference_value()`, `GradRuleset.reference_grad()`, and `pullback.apply()`.
> 
> I did not mean nested mixed spaces by "multiple meshes + multiple spaces". I just meant (meshA, elemA) x (meshB, elemB). Though we do not use nested mixed spaces in Firedrake, some functions such as `flatten_domain_element` are written with this in mind as you might have noticed. I do not think consideration of such nested mixed spaces would affect the mixed-problem design in UFL, though.
> 
> @jhale I think the use of `Mixed` in `MixedMesh` is consistent with the use of `Mixed` in `MixedElement`. If we consider assembling (i, j)-block of a matrix, we can view it as mixing `mixed_mesh[i]` and `mixed_mesh[j]` as well as mixing `mixed_element[i]` and `mixed_element[j]`.
> 
> But I agree that it is confusing. As @dham suggested, `MeshSequence` or `MultiMesh` is also fine by me.

Let's not use `MultiMesh`, as that has been used in FEniCS before to mean hierarchies of meshes.

jhale:
`MeshSequence` is clearer and at least to my understanding it is consistent with Python's abstract base class terminology `collections.abc.Sequence`.

@ksagiyam  I see your analogy between `MixedMesh` and `MixedElement` at the form level, however, `MixedElement` was named after a widely established and accepted terminology from the finite element literature. We don't have this for `MixedMesh`, and so my preference is to take precise terminology from elsewhere.

ufl/algorithms/analysis.py
ksagiyam: Done.

ufl/domain.py
ksagiyam: Fixed. `AbstractFiniteElement` declares `num_sub_elements`.

ufl/domain.py
```diff
@@ -190,26 +330,34 @@ def as_domain(domain):
     """Convert any valid object to an AbstractDomain type."""
     if isinstance(domain, AbstractDomain):
         # Modern UFL files and dolfin behaviour
+        (domain,) = set(domain.meshes)
```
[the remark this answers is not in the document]
ksagiyam: I think `as_domain` is more widely used. In Firedrake I can see `as_domain(V)` for instance.

ufl/domain.py
ksagiyam: Done.

ufl/algorithms/apply_derivatives.py
ksagiyam: Shape of `g` (which I now call `rgrad`) is slightly different for scalar and non-scalar spaces, e.g. `(3,)` v.s. `(5, 3)` for `topological dim = 3`, so `g[(dofoffset + idx,) + (j,)]` fails for scalar spaces. We could special case for the scalar space inside the `flatten_domain_element` loop, but I thought it would be cleaner if we just did that outside.

ufl/algorithms/apply_derivatives.py
ksagiyam: You are right. Simplified as suggested and added more comments. Thanks.

ufl/algorithms/apply_derivatives.py
ksagiyam: I think so.

ufl/algorithms/apply_derivatives.py
ksagiyam: Made similar changes here, too.

ufl/algorithms/compute_form_data.py
ksagiyam: Elaborated the error message.

ufl/domain.py
ksagiyam: Added type-hinting. Right, we now use `num_sub_elements`.

ufl/domain.py
```diff
@@ -126,6 +153,119 @@ def _ufl_sort_key_(self):
...
+        elem0 = FiniteElement("Lagrange", cell, 1, (), identity_pullback, H1)
+        elem1 = FiniteElement("Lagrange", cell, 2, (), identity_pullback, H1)
+        elem = MixedElement([elem0, elem1])
+        V = FunctionSpace(domain, elem)
+        v = TestFunction(V)
+        v0, v1 = split(v)
+
+    """
+
+    def __init__(self, meshes, ufl_id=None, cargo=None):
```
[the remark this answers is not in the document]
ksagiyam: Done.

ufl/domain.py
ksagiyam: Done.

ufl/domain.py
ksagiyam: Done.

ksagiyam:
Thanks, all. Yes, let us use `MeshSequence`. The followings are the changes I have made according to the suggestions:

- Addressed @jorgensd 's suggestions,
- Renamed `MixedMesh` `MeshSequence`,
- Removed `ufl_id` from `MeshSequence`; I realised that my Firedrake branch broke after a recent rebase, and found that it was caused by distinguishing two `MeshSequence` instances made from the same set of `Mesh`es. Thinking about this, I thought `ufl_id` in `MeshSequence` is redundant; please see updated `MeshSequence.__repr__()` and `MeshSequence._ufl_hash_data_()`. I also added a test to highlight this point.

@jhale What sort of tests do you think are missing? After this PR, I will soon make another PR that adds logics to handle restrictions (to handle cases where exterior facet integration of meshA is interior facet integration of meshB, etc). At that stage I will add one or two more examples that test facet integrations.

ksagiyam:
@jorgensd Do you have any further comments on this PR by any chance? I would like to have this PR go.

jorgensd:
> @jorgensd Do you have any further comments on this PR by any chance? I would like to have this PR go.

I’d like @garth-wells, @chrisrichardson and @michalhabera  to have a look over this. If they don’t have time, I guess we should get towards merging this.

ufl/domain.py
```diff
@@ -50,6 +58,33 @@ def topological_dimension(self):
...
+        """Return i-th component mesh."""
+        if i >= len(self):
+            raise ValueError(f"index ({i}) >= num. component meshes ({len(self)})")
+        return self.meshes[i]
+
+    def __iter__(self):
+        """Return iterable component meshes."""
+        return iter(self.meshes)
+
+    def iterable_like(self, element: AbstractFiniteElement) -> Iterable["Mesh"] | "MeshSequence":
```
mscroggs, line 80: Do we need to increase the minimum python version in pyproject.toml to  allow using `|` in typing?
Given the minimum version is currently 3.8 which has passed end-of-life, we should probably bump it anyway, but not necessarily in this PR unless it's needed here

mscroggs (approved):

ufl/domain.py
ksagiyam: Thanks. It's probably not a good idea to bump python version in this PR. I just switched to `typing.Union`.

jorgensd (approved):

PR #92, opened 2022-02-24
Group integrals with common integrand in IntegralData

jorgensd:
MWE:
```
from ufl import FiniteElement, TrialFunction, TestFunction, dx, Mesh, VectorElement, triangle, FunctionSpace, inner, grad
import ufl.algorithms
import ufl.classes
mesh = Mesh(VectorElement("Lagrange", triangle, 1))
V = FunctionSpace(mesh, FiniteElement("Lagrange", triangle, 1))
u = TrialFunction(V)
v = TestFunction(V)
a = inner(u, v)*dx((1, 2, 3)) + inner(grad(u), grad(v))*dx(2) + inner(u, v)*dx

form_data = ufl.algorithms.compute_form_data(
    a,
    do_apply_function_pullbacks=True,
    do_apply_integral_scaling=True,
    do_apply_geometry_lowering=True,
    preserve_geometry_types=(ufl.classes.Jacobian,),
    do_apply_restrictions=True,
    do_append_everywhere_integrals=False)
for integral in form_data.integral_data:
    print(str(integral))
```
On this branch, we get two `IntegralData`, one for `inner(u,v)*dx((1,3,"everywhere"))` and one for `(inner(u,v) + inner(grad(u), grad(v)))*dx((2,))`.
On main one would get four integral datas; `inner(u,v)*dx(i)` for `i=1,3,"everywhere"` and `(inner(u,v) + inner(grad(u), grad(v)))*dx((2,))`.

This is relevant when loading complex meshes with hundreds of subdomain_ids, where a tuple of subdomain ids is used to indicate a union of domains to integrate over. This would cause the compilers in `Firedrake` and `DOLFINx` to generate hundreds of duplicate kernels, which makes the generated code unreadable and slow to generate.

PR #305, opened 2024-09-19
Move grouping of integral from `build_integral_data` to `group_form_integrals`

### LINKED ISSUE #304: Uniqueness of integrands after computing form data
Using `compute_form_data` makes equal integrands un-equal in ufl.
MWE
```python
from ufl import TestFunction, Mesh, interval, FunctionSpace
import ufl.algorithms
import ufl.classes
cell = interval
mesh = Mesh(FiniteElement("Lagrange", cell, 1, (2, ), identity_pullback, H1))
V = FunctionSpace(mesh, FiniteElement("Lagrange", cell, 1, (), identity_pullback, H1))
u = TestFunction(V)
a0 = u.dx(0) * ufl.dx((1))
a1 = u.dx(0) * ufl.dx((2))

print(a0.integrals()[0].integrand() == a1.integrals()[0].integrand())
form_data0 = ufl.algorithms.compute_form_data(
    a0+a1,
    do_apply_function_pullbacks=True,
    do_apply_integral_scaling=True,
    do_apply_geometry_lowering=True,
    preserve_geometry_types=(ufl.classes.Jacobian,),
    do_apply_restrictions=True,
    do_append_everywhere_integrals=False)
i0 = form_data0.integral_data[0].integrals[0].integrand()
i1 = form_data0.integral_data[1].integrals[0].integrand()
print(str(i0))
print(str(i1))
print(i0 == i1)
```
Returning
```bash
True
({ A | A_{i_{12}} = sum_{i_{13}} ({ A | A_{i_{17}, i_{18}} = ({ A | A_{i_{15}, i_{14}} = J[i_{14}, i_{15}] })[i_{17}, i_{18}] / (sum_{i_{16}} J[i_{16}, 0] * J[i_{16}, 0] ) })[i_{13}, i_{12}] * (reference_grad(reference_value(v_0)))[i_{13}]  })[0] * weight * |(cell_orientation * sqrt(sum_{i_{10}} J[i_{10}, 0] * J[i_{10}, 0] ))|
({ A | A_{i_{12}} = sum_{i_{13}} ({ A | A_{i_{22}, i_{23}} = ({ A | A_{i_{20}, i_{19}} = J[i_{19}, i_{20}] })[i_{22}, i_{23}] / (sum_{i_{21}} J[i_{21}, 0] * J[i_{21}, 0] ) })[i_{13}, i_{12}] * (reference_grad(reference_value(v_0)))[i_{13}]  })[0] * weight * |(cell_orientation * sqrt(sum_{i_{11}} J[i_{11}, 0] * J[i_{11}, 0] ))|
False
```

Any idea how we can resolve this, as it results in duplicate kernels similar to what I tried to resolve with: https://github.com/FEniCS/ufl/pull/92

jorgensd:
Move group to resolve #304 by grouping prior to pull-back.
This will resolve https://github.com/FEniCS/ffcx/issues/716 as well.

I think this is a better fix than what I came up with in https://github.com/FEniCS/ufl/pull/92/files
as we now group the integrals in the function that is supposed to do grouping, and it will reduce the load of the other functions (pull-back, geometry_lowering etc.) since it happens at the start of `compute_form_data`.

### DIFF
--- a/ufl/algorithms/domain_analysis.py
+++ b/ufl/algorithms/domain_analysis.py
@@ -15,6 +15,7 @@ from ufl.algorithms.coordinate_derivative_helpers import (
     attach_coordinate_derivatives,
     strip_coordinate_derivatives,
 )
+from ufl.algorithms.renumbering import renumber_indices
 from ufl.form import Form
 from ufl.integral import Integral
 from ufl.protocols import id_or_none
@@ -372,7 +373,7 @@ def group_form_integrals(form, domains, do_append_everywhere_integrals=True):
         meta_hash = hash(canonicalize_metadata(metadata))
         subdomain_id = integral.subdomain_id()
         subdomain_data = id_or_none(integral.subdomain_data())
-        integrand = integral.integrand()
+        integrand = renumber_indices(integral.integrand())
         unique_integrals[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] += (
             subdomain_id,
         )
--- a/ufl/algorithms/domain_analysis.py
+++ b/ufl/algorithms/domain_analysis.py
@@ -262,37 +262,17 @@ def build_integral_data(integrals):
     itgs = defaultdict(list)
 
     # --- Merge integral data that has the same integrals,
-    unique_integrals = defaultdict(tuple)
-    metadata_table = defaultdict(dict)
     for integral in integrals:
-        integrand = integral.integrand()
         integral_type = integral.integral_type()
         ufl_domain = integral.ufl_domain()
-        metadata = integral.metadata()
-        meta_hash = hash(canonicalize_metadata(metadata))
-        subdomain_id = integral.subdomain_id()
-        subdomain_data = id_or_none(integral.subdomain_data())
-        if subdomain_id == "everywhere":
+        subdomain_ids = integral.subdomain_id()
+        if "everywhere" in subdomain_ids:
+ 
             raise ValueError(
                 "'everywhere' not a valid subdomain id. "
                 "Did you forget to call group_form_integrals?"
             )
-        unique_integrals[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] += (
-            subdomain_id,
-        )
-        metadata_table[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] = metadata
-
-    for integral_data, subdomain_ids in unique_integrals.items():
-        (integral_type, ufl_domain, metadata, integrand, subdomain_data) = integral_data
-
-        integral = Integral(
-            integrand,
-            integral_type,
-            ufl_domain,
-            subdomain_ids,
-            metadata_table[integral_data],
-            subdomain_data,
-        )
+        
         # Group for integral data (One integral data object for all
         # integrals with same domain, itype, (but possibly different metadata).
         itgs[(ufl_domain, integral_type, subdomain_ids)].append(integral)
@@ -380,7 +360,30 @@ def group_form_integrals(form, domains, do_append_everywhere_integrals=True):
                         )
                         integral = attach_coordinate_derivatives(integral, samecd_integrals[0])
                         integrals.append(integral)
-    return Form(integrals)
+
+    # Group integrals by common subdomain id
+    # u.dx(0)*dx(1) + u.dx(0)*dx(2) -> u.dx(0)*dx((1,2))
+    # to avoid duplicate kernels generated after geoemtry lowering
+    unique_integrals = defaultdict(tuple)
+    metadata_table = defaultdict(dict)
+    for integral in integrals:
+        integral_type = integral.integral_type()
+        ufl_domain = integral.ufl_domain()
+        metadata = integral.metadata()
+        meta_hash = hash(canonicalize_metadata(metadata))
+        subdomain_id = integral.subdomain_id()
+        subdomain_data = id_or_none(integral.subdomain_data())
+        unique_integrals[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] += (subdomain_id,)
+        metadata_table[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] = metadata
+    
+    grouped_integrals = []
+    for integral_data, subdomain_ids in unique_integrals.items():
+        (integral_type, ufl_domain, metadata, integrand, subdomain_data) = integral_data
+        integral = Integral(integrand, integral_type, ufl_domain, subdomain_ids,
+                            metadata_table[integral_data], subdomain_data)
+        grouped_integrals.append(integral)
+
+    return Form(grouped_integrals)
 
 
 def reconstruct_form_from_integral_data(integral_data):
--- a/ufl/algorithms/domain_analysis.py
+++ b/ufl/algorithms/domain_analysis.py
@@ -372,6 +372,7 @@ def group_form_integrals(form, domains, do_append_everywhere_integrals=True):
         meta_hash = hash(canonicalize_metadata(metadata))
         subdomain_id = integral.subdomain_id()
         subdomain_data = id_or_none(integral.subdomain_data())
+        integrand = integral.integrand()
         unique_integrals[(integral_type, ufl_domain, meta_hash, integrand, subdomain_data)] += (
             subdomain_id,
         )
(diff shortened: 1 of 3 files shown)


jorgensd:
Seems like one should use the `IndexRenumberingTransform`: https://github.com/FEniCS/ufl/blob/004a678320e5c65015d56071b968ee08314aecc1/ufl/algorithms/renumbering.py#L35-L39
@wence- do you have any other ideas?

ufl/algorithms/domain_analysis.py
```diff
@@ -380,7 +360,39 @@ def calc_hash(cd):
...
+    unique_integrals = defaultdict(tuple)
+    metadata_table = defaultdict(dict)
+    for integral in integrals:
+        integral_type = integral.integral_type()
+        ufl_domain = integral.ufl_domain()
+        metadata = integral.metadata()
+        meta_hash = hash(canonicalize_metadata(metadata))
+        subdomain_id = integral.subdomain_id()
+        subdomain_data = id_or_none(integral.subdomain_data())
+        integrand = renumber_indices(integral.integrand())
```
mscroggs, line 376: This line is where the function that warns that it's badly designed is used

wence-:
> Seems like one should use the `IndexRenumberingTransform`:
> 
> https://github.com/FEniCS/ufl/blob/004a678320e5c65015d56071b968ee08314aecc1/ufl/algorithms/renumbering.py#L35-L39
> 
> 
> @wence- do you have any ideas on how to improve it?

It doesn't look that bad. I would forward-port the model to the `MultiFunction` case and use `map_expr_dags` instead.

What you want to know is "are two dags equal to each other, up to index relabelling?". This is achievable by visiting the dag, child before parent, and replacing free indices with a unique new index.

```
class IndexRelabeller(MultiFunction):
    def __init__(self):
        super().__init__()
        count = itertools.count()
        self.index_cache = defaultdict(lambda: Index(next(count)))

    expr = MultiFunction.reuse_if_untouched
   
    def multi_index(self, o):
        return type(o)(tuple(self.index_cache[i] if isinstance(i, Index) else i for i in o.indices()))

    def zero(self, o):
        fi = o.ufl_free_indices
        fid = o.ufl_free_index_dimensions
        new_indices = [self.index_cache[Index(i)].count() for i in fi]
        new_fi, new_fid = zip(*sorted(zip(new_indices, fid), key=lambda x: x[0]))
        return type(o)(o.ufl_shape, tuple(new_fi), tuple(new_fid))

map_expr_dags(IndexRelabeller(), whatever)
```
Would probably work, untested.

jorgensd:
> > Seems like one should use the `IndexRenumberingTransform`:
> > https://github.com/FEniCS/ufl/blob/004a678320e5c65015d56071b968ee08314aecc1/ufl/algorithms/renumbering.py#L35-L39
> > 
> > @wence- do you have any ideas on how to improve it?
> 
> It doesn't look that bad. I would forward-port the model to the `MultiFunction` case and use `map_expr_dags` instead.
> 
> What you want to know is "are two dags equal to each other, up to index relabelling?". This is achievable by visiting the dag, child before parent, and replacing free indices with a unique new index.
> 
> ```
> class IndexRelabeller(MultiFunction):
>     def __init__(self):
>         super().__init__()
>         count = itertools.count()
>         self.index_cache = defaultdict(lambda: Index(next(count)))
> 
>     expr = MultiFunction.reuse_if_untouched
>    
>     def multi_index(self, o):
>         return type(o)(tuple(self.index_cache[i] if isinstance(i, Index) else i for i in o.indices()))
> 
>     def zero(self, o):
>         fi = o.ufl_free_indices
>         fid = o.ufl_free_index_dimensions
>         new_indices = [self.index_cache[Index(i)].count() for i in fi]
>         new_fi, new_fid = zip(*sorted(zip(new_indices, fid), key=lambda x: x[0]))
>         return type(o)(o.ufl_shape, tuple(new_fi), tuple(new_fid))
> 
> map_expr_dags(IndexRelabeller(), whatever)
> ```
> 
> Would probably work, untested.

Thanks Lawrence! I will test this tomorrow!

ufl/algorithms/domain_analysis.py
wence-: Just to note, if you compute the `ufl_signature` of a form, then that does include an index renumbering, and so that might be another place to look for inspiration in how to do this.

ufl/algorithms/renumbering.py
```diff
@@ -1,87 +1,71 @@
 """Algorithms for renumbering of counted objects, currently variables and indices."""
-# Copyright (C) 2008-2016 Martin Sandve Alnæs and Anders Logg
+# Copyright (C) 2008-2024 Martin Sandve Alnæs, Anders Logg, Jø¶gen S. Dokken and Lawerence Mitchell
```
jorgensd, line 2:
```suggestion
# Copyright (C) 2008-2024 Martin Sandve Alnæs, Anders Logg, Jørgen S. Dokken and Lawerence Mitchell
```

wence-: That's not how my name is spelt, but I don't need it to be here either

ufl/algorithms/renumbering.py
```diff
@@ -1,87 +1,71 @@
 """Algorithms for renumbering of counted objects, currently variables and indices."""
-# Copyright (C) 2008-2016 Martin Sandve Alnæs and Anders Logg
+# Copyright (C) 2008-2024 Martin Sandve Alnæs, Anders Logg, Jørgen S. Dokken and Lawerence Mitchell
```
jorgensd, line 2:
```suggestion
# Copyright (C) 2008-2024 Martin Sandve Alnæs, Anders Logg, Jørgen S. Dokken and Lawrence Mitchell
```

ufl/algorithms/renumbering.py
jorgensd: Fixed now, sorry about that. I think you should be there as you wrote 99 % of that code (up to a typo).

ufl/algorithms/renumbering.py
```diff
@@ -1,87 +1,71 @@
...
+    elif isinstance(form, Integral):
+        integral = form
+        reindexer = IndexRelabeller()
+        new_integrand = map_expr_dag(reindexer, integral.integrand())
+        return integral.reconstruct(new_integrand)
+    elif isinstance(form, Expr):
+        expr = form
+        reindexer = IndexRelabeller()
+        return map_expr_dag(reindexer, expr)
+    else:
```
wence-, line 70: Is this not exactly what `map_integrand_dags` is for? I forget.

I'll note that the index renumbering pass here is still a "bad" algorithm, though it is at least dag aware. In the following way: index numbers only matter for free indices, not bound indices, because `A = ComponentTensor(A[i, j], (i, j))` and `B = ComponentTensor(A[k, l], (k, l))` are identical expressions. It might be that what you want this for, this is not a problem. There is some discussion of this in section 9.2.5 of Miklos' thesis: https://spiral.imperial.ac.uk/bitstream/10044/1/62636/1/Homolya-M-2018-PhD-Thesis.pdf

jorgensd: I am not sure I follow your first comment here.
I am following the same logic as in:
https://github.com/FEniCS/ufl/blob/60a6956a553a014d9b9dc5fbd66c7e764dea0f72/ufl/algorithms/apply_geometry_lowering.py#L473-L497

For the second comment: For the issue at hand, I've only encountered issues with free indices, so I am not concerned with bound indices for now.

garth-wells:
@jorgensd is this ready, or does it need more feedback?

ufl/algorithms/renumbering.py
wence-: > I am not sure I follow your first comment here.

https://github.com/FEniCS/ufl/blob/main/ufl/algorithms/map_integrands.py

`map_integrands` (called via `map_integrand_dags`) does (in a more systematic way) this handling of different types of input.

wence-: It doesn't really matter I suppose.

jorgensd:
> @jorgensd is this ready, or does it need more feedback?

I believe this is ready for merging. @jpdean have you tested it for the use-cases that spawned revisiting this issue?

ufl/algorithms/renumbering.py
jorgensd: It does look a whole lot neater with `map_integrand_dags`!

PR #307, opened 2024-09-20
Move FiniteElement.components to FunctionSpace.components

mscroggs:
Docstring is about a physical cell which the finite element is no longer aware of, but the function space is

### DIFF
--- a/ufl/functionspace.py
+++ b/ufl/functionspace.py
@@ -60,6 +60,32 @@ class BaseFunctionSpace(AbstractFunctionSpace, UFLObject):
         self._ufl_domain = domain
         self._ufl_element = element
 
+    @property
+    def components(self) -> typing.Dict[typing.Tuple[int, ...], int]:
+        """Get the numbering of the components of the element of this space.
+
+        Returns:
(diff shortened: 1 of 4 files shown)


ufl/functionspace.py
```diff
@@ -60,6 +62,32 @@ def __init__(self, domain, element, label=""):
...
 
+    @property
+    def components(self) -> typing.Dict[typing.Tuple[int, ...], int]:
+        """Get the numbering of the components of the element of this space.
+
+        Returns:
+            A map from the components of the values on a physical cell (eg (0, 1))
+            to flat component numbers on the reference cell (eg 1)
+        """
+        from ufl.pullback import SymmetricPullback
```
jorgensd, line 73: Can we move the import outside the function itself?

jorgensd: Then I am fine with this.

jorgensd:
@dham do you have any objections to merging this?

dham:
I think this is a natural consequence of moving geometric quantities out of elements. I have no objection.

PR #308, opened 2024-09-20
Fix extract block for tensor spaces

jorgensd:
To be able to easily extract blocks from `ufl.MixedFuntionSpaces`.

Motivation:
Instead of writing out every block of `a` and `L` by hand, as
```python
    a00 = ufl.inner(ufl.grad(u), ufl.grad(v)) * dx
    a01 = ufl.inner(c, v) * dx
    a10 = ufl.inner(u, d) * dx

    L0 = ufl.inner(f, v) * dx + ufl.inner(g, v) * ufl.ds
    L1 = ufl.inner(zero, d) * dx

    a = dolfinx.fem.form([[a00, a01], [a10, None]], dtype=dtype)
    L = dolfinx.fem.form([L0, L1], dtype=dtype)

```
one can now do
```python
    dx = ufl.Measure("dx", domain=mesh)
    f = -ufl.div(ufl.grad(u_ex))
    n = ufl.FacetNormal(mesh)
    g = ufl.dot(ufl.grad(u_ex), n)
    a = ufl.inner(ufl.grad(u), ufl.grad(v)) * dx
    a += ufl.inner(c, v) * dx
    a += ufl.inner(u, d) * dx
    L = ufl.inner(f, v) * dx + ufl.inner(g, v) * ufl.ds
    L += ufl.inner(zero, d) * dx

    a_blocked = ufl.extract_blocks(a)
    L_blocked = ufl.extract_blocks(L)

    a = dolfinx.fem.form(a_blocked, dtype=dtype)
    L = dolfinx.fem.form(L_blocked, dtype=dtype)
```

It also fixes `extract_blocks` for tensors, and properly document this function, as to what happens if `i` or `j` is not provided.
Now it extracts the blocks as a scalar, vector or tensor.

### DIFF
--- a/ufl/algorithms/formsplitter.py
+++ b/ufl/algorithms/formsplitter.py
@@ -96,7 +96,7 @@ def extract_blocks(form, i: Optional[int] = None, j: Optional[None] = None):
     Args:
         form: A form
         i: Index of the block to extract. If set to ``None``, ``j`` must be None.
-        j: Index of the block to extract. If set to ``None``, ``i`` must be None.
+        j: Index of the block to extract.
     """
     fs = FormSplitter()
     arguments = form.arguments()
--- a/ufl/algorithms/formsplitter.py
+++ b/ufl/algorithms/formsplitter.py
@@ -8,6 +8,8 @@
 #
 # Modified by Cecile Daversin-Catty, 2018
 
+from typing import Optional
+
 from ufl.algorithms.map_integrands import map_integrand_dags
 from ufl.argument import Argument
 from ufl.classes import FixedIndex, ListTensor
@@ -31,20 +33,10 @@ class FormSplitter(MultiFunction):
         if obj.part() is not None:
             # Mixed element built from MixedFunctionSpace,
             # whose sub-function spaces are indexed by obj.part()
-            if len(obj.ufl_shape) == 0:
-                if obj.part() == self.idx[obj.number()]:
-                    return obj
-                else:
-                    return Zero()
+            if obj.part() == self.idx[obj.number()]:
+                return obj
             else:
-                indices = [()]
-                for m in obj.ufl_shape:
-                    indices = [(k + (j,)) for k in indices for j in range(m)]
-
-                if obj.part() == self.idx[obj.number()]:
-                    return as_vector([obj[j] for j in indices])
-                else:
-                    return as_vector([Zero() for j in indices])
+                return Zero(obj.ufl_shape)
         else:
             # Mixed element built from MixedElement,
             # whose sub-elements need their function space to be created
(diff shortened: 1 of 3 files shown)


ufl/finiteelement.py
```diff
@@ -148,7 +148,7 @@ def components(self) -> _typing.Dict[_typing.Tuple[int, ...], int]:
         offset = 0
         c_offset = 0
         for e in self.sub_elements:
-            for i, j in enumerate(np.ndindex(e.value_shape)):
+            for i, j in enumerate(np.ndindex(e.reference_value_shape)):
```
mscroggs, line 151: This change shouldn't be in this PR
```suggestion
            for i, j in enumerate(np.ndindex(e.value_shape)):
```

mscroggs (approved):

ufl/finiteelement.py
jorgensd: I'll revert it once 307 is merged.

ufl/algorithms/formsplitter.py
```diff
@@ -92,33 +84,60 @@ def multi_index(self, obj):
...
+    If arity is 0, returns the form.
+    If arity is 1, return the ith block. If ``i`` is ``None``, return all blocks.
+    If arity is 2, return the ``(i,j)`` entry. If ``j`` is ``None``, return the ith row.
+
+    If neither `i` nor `j` are set, return all blocks (as a scalar, vector or tensor).
+
+    Args:
+        form: A form
+        i: Index of the block to extract. If set to ``None``, ``j`` must be None.
+        j: Index of the block to extract. If set to ``None``, ``i`` must be None.
```
cdaversin, line 99: Not consistent with : "If ``j`` is ``None``, return the ith row."

jorgensd, line 99:
```suggestion
        j: Index of the block to extract.
```

ufl/algorithms/formsplitter.py
```diff
@@ -131,10 +150,9 @@ def extract_blocks(form, i=None, j=None):
     except TypeError:
         # Only one form returned
         forms_tuple = (forms,)
-
     if i is not None:
         if arity > 1 and j is not None:
-            return forms_tuple[i * len(parts) + j]
+            return forms_tuple[i][j]
         else:
             return forms_tuple[i]
```
cdaversin, line 157: Might be useful to have a warning if we end up there with j != None

ufl/algorithms/formsplitter.py
```diff
@@ -92,33 +84,60 @@ def multi_index(self, obj):
...
+                forms.append(tuple(form_i))
+            return tuple(forms)
         else:
-            # If standard element, extract only part
-            parts.append(a.part())
-    parts = tuple(sorted(set(parts)))
-    for pi in parts:
+            return fs.split(form, i, j)
+
+    # If mixed function space, each argument has sub-elements
```
cdaversin, line 128: Maybe some safety check on i and j depending on number of components in the MixedFunctionSpace ?

cdaversin (approved):
Looks good to me :)

jpdean:
Nice! This'll be really handy for coupled problems. Do we have a test?

jorgensd:
> Nice! This'll be really handy for coupled problems. Do we have a test?

I think we can adapt the stokes demo.

I use this in many of my own works lately.

PR #330, opened 2024-12-05
Introduce ridge integrals

jorgensd:
Continuation of @cdaversin work on 3D-1D coupling and 2D-0D


.github/workflows/fenicsx-tests.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: main
+          ref: jpdean/multiple_subdomain_data
```
jorgensd, line 80:
```suggestion
          ref: main
```

ufl/measure.py
```diff
@@ -32,6 +32,8 @@
     # === Integration over topological dimension - 1:
     ("exterior_facet", "ds"),  # Over one-sided exterior facets of a mesh
     ("interior_facet", "dS"),  # Over two-sided facets between pairs of adjacent cells of a mesh
+    # === Integration over topological dimension - 2:
+    ("edge", "dl"),  # Over edges of a mesh
```
mscroggs, line 36: edges are only tdim - 2 in a 3d mesh. Should these be called "ridges" instead (See eg https://en.wikipedia.org/wiki/Face_(geometry)#k-face)

dham: Good catch! Presumably `dl` integrates over points in 2d (just like `ds` and `dS` in 1D).

AUTHORS
```diff
@@ -28,7 +28,8 @@ Contributors:
   | Corrado Maurini       <corrado.maurini@upmc.fr>
   | Jack S. Hale          <mail@jackhale.co.uk>
   | Tuomas Airaksinen     <tuomas.airaksinen@gmail.com>
-  | Nacime Bouziani       <nacime.bouziani@gmail.com>
```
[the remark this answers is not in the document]
mscroggs: I almost commented this too but then spotted a duplicate Nacime further down the list

ufl/measure.py
jorgensd: > ridge

I was thinking about this when looking through UFL. I guess we should decide if we want to support "Edge" (i.e. only 3D-1D), or ridge (2D-0D, 3D-1D)? @cdaversin do you have any opinions?

dham: I think this is conflating "things that are expressible in UFL" with "things that are supported in Dolfinx". Lots of things can be expressed in UFL that individual solvers don't support. The point is that the UFL notation should match the maths as closely as possible so that we don't have to go back and make a backward incompatible change when someone expands their capabilities in the future.

I think that the UFL feature should be codimension 2, even if that's not what is supported in Dolfinx right now.

AUTHORS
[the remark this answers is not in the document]
nbouziani: `n.bouziani18@imperial.ac.uk` can be kept.

jorgensd:
I’ve changed the relevant classes to `ridge`

HoseongJeong:
I truly appreciate dokken.

mscroggs (approved):

PR #331, opened 2024-12-12
Fix action space check

pbrubeck:
The space check in `action` was not quite right, it is hard to tell exactly why. Previously it only passed because `V == V.dual()` in Firedrake.

This PR removes the need to check specific types of the arguments and handles them in a more symmetric way.


pbrubeck:
This ruff thing is ridiculous

ufl/action.py
```diff
@@ -159,29 +157,36 @@ def __hash__(self):
 
 def _check_function_spaces(left, right):
```
pbrubeck, line 158: Dear code reviewers,

Please ignore the rest of the diff, that's just ruff.
Only take a look at this function.

Thank you.

pbrubeck:
@jorgensd FeniCSx CI fails, but I suspect it is for an unrelated reason?

jorgensd:
> @jorgensd FeniCSx CI fails, but I suspect it is for an unrelated reason?

@finsberg, it seems like you missed a test with your latest PR:
https://github.com/FEniCS/dolfinx/blob/0c3a33ec5cfc5fef3e5bc965fcd3df533ed4aee6/python/test/unit/io/test_xdmf_function.py#L232-L249

dham (approved):
I think this looks right.

PR #335, opened 2024-12-21
Fix FormSum weights

pbrubeck:
FormSum weights were left out in most parts of the code.


dham (changes requested):
Please to be adding tests

dham (approved):

jorgensd (approved):

PR #336, opened 2024-12-30
Simplify indexed ListTensor objects

pbrubeck:
When splitting mixed Forms, mixed Arguments get replaced by ListTensors of Zero and subspace Arguments of the form `[0,  ..., 0, v[0], ..., v[k],  0, ..., 0]`. The resulting split Form would contain many instances of Indexed objects with this ListTensor, that could be simplified to either `0` or `[v[0], ..., v[k]] -> v`. 

Previously, the simplification step was left as a task for the form compiler, but working with such complicated expressions often results in sub-optimal code generation. This PR addresses the following simplications at the UFL-level:

1. `Indexed(ListTensor[v0, ..., vk], (j,)) -> vj`
2. `ListTensor(Indexed(v, FixedIndex(i)) for i in range(k)) -> v`
3. `Restricted(ConstantValue) -> ConstantValue`

As a consequence, splitting a bilinear Form with a block in the diagonal equal to zero produces an empty Form, losing all the information that the unsimplified Form carried around, such as the Arguments and the arity. This situation needs to be handled carefully as a special case. See https://github.com/firedrakeproject/firedrake/pull/3947

With these simplifications, we can correctly estimate quadrature degrees for integrands that only depend on one component of the solution, thus lowering flop counts.

### DIFF
--- a/ufl/indexed.py
+++ b/ufl/indexed.py
@@ -47,7 +47,8 @@ class Indexed(Operator):
 
         try:
             # Simplify indexed ListTensor
-            return expression[multiindex]
+            c = expression[multiindex]
+            return Indexed(*c.ufl_operands) if isinstance(c, Indexed) else c
         except ValueError:
             return Operator.__new__(cls)
 
--- a/ufl/indexed.py
+++ b/ufl/indexed.py
@@ -26,15 +26,30 @@ class Indexed(Operator):
 
     def __new__(cls, expression, multiindex):
         """Create a new Indexed."""
+        # cyclic import
         from ufl.tensors import ListTensor
 
+        simpler = False
         indices = multiindex.indices()
-        while isinstance(expression, ListTensor) and isinstance(indices[0], FixedIndex):
+        while (
+            len(indices) > 0
+            and isinstance(expression, ListTensor)
+            and isinstance(indices[0], FixedIndex)
+        ):
             # Simplify indexed ListTensor objects
             expression = expression[indices[0]]
             indices = indices[1:]
+            simpler = True
 
-        if isinstance(expression, Zero):
+        if isinstance(expression, Indexed):
+            # Simplify nested Indexed objects
+            indices = expression.ufl_operands[1].indices() + indices
+            expression = expression.ufl_operands[0]
+            simpler = True
+
+        if len(indices) == 0:
+            return expression
+        elif isinstance(expression, Zero):
             # Zero-simplify indexed Zero objects
             shape = expression.ufl_shape
             efi = expression.ufl_free_indices
@@ -49,8 +64,8 @@ class Indexed(Operator):
             else:
                 fi, fid = (), ()
             return Zero(shape=(), free_indices=fi, index_dimensions=fid)
-        elif expression.ufl_shape == () and multiindex == ():
-            return expression
+        elif simpler:
+            return Indexed(expression, MultiIndex(indices))
         else:
             return Operator.__new__(cls)
 
--- a/ufl/indexed.py
+++ b/ufl/indexed.py
@@ -26,8 +26,7 @@ class Indexed(Operator):
 
     def __new__(cls, expression, multiindex):
         """Create a new Indexed."""
-        indices = multiindex.indices()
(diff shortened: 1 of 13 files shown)


jhale:
This seems fine to me and thanks for the various modernisations added as well.

ufl/indexed.py
```diff
@@ -41,9 +43,12 @@ def __new__(cls, expression, multiindex):
...
                 fi, fid = (), ()
             return Zero(shape=(), free_indices=fi, index_dimensions=fid)
-        elif expression.ufl_shape == () and multiindex == ():
-            return expression
-        else:
+
+        try:
+            # Simplify indexed ListTensor
+            c = expression[multiindex]
+            return Indexed(*c.ufl_operands) if isinstance(c, Indexed) else c
```
ksagiyam, line 50: Can we not do:
```suggestion
            return expression[multiindex]
```
?

pbrubeck: If `__new__` returns an old `Indexed` object this causes an infinite recursion. Rebuilding the object seems to avoid this issue.

pbrubeck: According to the [python documentation](https://docs.python.org/3/reference/datamodel.html#basic-customization):
> If `__new__()` returns an instance of cls, then the new instance’s `__init__()` method will be invoked like `__init__(self[, ...])`, where self is the new instance and the remaining arguments are the same as were passed to `__new__()`.

pbrubeck: Another way to prevent `Indexed.__init__()` to be called on already initialized instances is simply by leaving the `Indexed.__init__()` method unimplemented, and moving its contents to `Indexed.__new__()`.

ksagiyam: Is the `__new__() / __init__()` business the cause of infinite recursion? I can see that infinite recursion could be possible, though. Do you have an example which would cause infinite recursion to help me understand?

pbrubeck: That's right. You can use the code you are suggesting on this firedrake test to replicate the infinite recursion: `tests/firedrake/equation_bcs/test_equation_bcs.py::test_EquationBC_mixedpoisson_matrix[nonlinear]`

pbrubeck: I resolved this by introducing an `_init()` method called only on new instances within `__new__()`, in the same way other `Operator` subclasses like `Sum` do this.

ufl/indexed.py
```diff
@@ -89,6 +96,10 @@ def __init__(self, expression, multiindex):
         self.ufl_free_indices = fi
         self.ufl_index_dimensions = fid
 
+    def __init__(self, expression, multiindex):
+        """Initialise."""
+        Operator.__init__(self)
+
```
ksagiyam, line 102: So you do not pass `expression, multiindex` to avoid infinite recursion?

I think a more natural way to avoid this `__new__() / __init__()` game is to attach `_initialised` property to the object, and let `__init__()` immediately return if `self._initialised`. This way you can avoid making a new instance.

pbrubeck: If the object is unitialized, at which point are we supposed to attach this property?

pbrubeck: What about?
```suggestion
    def __init__(self, expression, multiindex):
        """Initialise."""
        if hasattr(self, "ufl_operands"):
        # This instance has been initialized before
            return
        # Store operands
        Operator.__init__(self, (expression, multiindex))
```

ksagiyam: I would first revert `_init()`, and do something like:

```
def __new__():

    self = Operator.__new__(cls)
    self._initialised = False

def __init__(self, ...)
    if self._initialised:
        return
    ...
    self._initialised = True

```
See also[](https://github.com/firedrakeproject/firedrake/blob/723ca723657332963b4f583a2772e44b07ce1781/firedrake/assemble.py#L917).

pbrubeck: Right. I need to add "_initialised" to the `__slots__` of the class.

ksagiyam (approved):

PR #338, opened 2025-01-16
Indexed: avoid contraction of repeated indices

pbrubeck:


ufl/indexed.py
```diff
@@ -45,13 +48,27 @@ def __new__(cls, expression, multiindex):
...
+
+                kk = list(kk)
+                if all(j in kk for j in jj):
+                    rep = dict(zip(jj, ii))
+                    Cind = tuple(rep.get(k, k) for k in kk)
+                    return Indexed(C, MultiIndex(Cind))
+
+        self = Operator.__new__(cls)
+        self._initialised = False
+        return self
```
ksagiyam, line 71: This might be a bit tricky, but I still prefer doing this by inheritance.

```
class Indexed():

    def __new__(cls, ...):
        ...
        try:
            # Simplification might be possible.
            return expression.index_and_simplify(multiindex)
        except NotImplementedError:
            # Index normally.
            self = Operator.__new__(cls)
            self._initialised = False
            return self
```

```
class Expr:
    ...
    def index_and_simplify(self, multiindex):
        raise NotImplementedErrror
    ...
```
```
class ListTensor:
    ...
    def index_and_simplify(self, multiindex):
        if isinstance(multiindex[0], FixedIndex):
            return Indexed(..., multiindex[1:])
        else:
            raise NotImplementedError
        ...
```
```
class ComponentTensor:
    ...
    def index_and_simplify(self, multiindex):
        ...
        return Indexed(..., ...)
```
This will make `Indexed` more compact; we should note that the treatment of `ComponentTensor` can be more general (`B` can be any scalar expression), in which case there would be more lines of code.

ksagiyam (changes requested):
Please add tests.

ufl/indexed.py
pbrubeck: Could you elaborate further on the treatment of `ComponentTensor`?

ksagiyam: For instance:
```
Indexed(
    ComponentTensor(
        Indexed(o1, (i, j)) + Indexed(o2, (i, j)),  # This is B.
        (i, j)
    ),
    (k, l)
)
```
should reduce to:
```
Indexed(o1, (k, l)) + Indexed(o2, (k, l)).
```
We probably need a DAG traversal on `B` that replaces indices in general. I have done that in in my PR https://github.com/firedrakeproject/ufl/blob/988189f8082646159d104638b966e0e779c46638/ufl/algorithms/apply_coefficient_split.py#L179, but, if we did it here, I would probably be able to remove `remove_component_and_list_tensors` in my PR.
I did not suggest that you generalise the treatment of ComponentTensor in this PR, though. My comment was just about the design.

ufl/tensors.py
```diff
@@ -120,6 +126,14 @@ def evaluate(self, x, mapping, component, index_values, derivatives=()):
         else:
             return a.evaluate(x, mapping, component, index_values)
 
+    def _simplify_indexed(self, multiindex):
+        """Return a simplified Expr used in the constructor of Indexed(self, multiindex)."""
+        k = multiindex[0]
+        if isinstance(k, FixedIndex):
+            sub = self.ufl_operands[int(k)]
+            return sub if len(multiindex) == 1 else Indexed(sub, MultiIndex(multiindex[1:]))
```
ksagiyam, line 134:
```suggestion
            return Indexed(sub, MultiIndex(multiindex[1:]))
```
Indexed.__new__() will handle this.

ksagiyam (changes requested):

ksagiyam:
You can add the following test:

```
from ufl import Index, triangle
from ufl.classes import Coefficient, ComponentTensor, FixedIndex, FunctionSpace, Indexed, ListTensor, Mesh, MultiIndex, Zero


def test_simplify_indexed():
    element = FiniteElement("Lagrange", triangle, 1, (3,), identity_pullback, H1)
    domain = Mesh(FiniteElement("Lagrange", triangle, 1, (2,), identity_pullback, H1))
    space = FunctionSpace(domain, element)
    u = Coefficient(space)
    z = Zero(())
    i = Index()
    j = Index()
    # ListTensor
    l = ListTensor(z, z, u[1])
    assert Indexed(l, MultiIndex((FixedIndex(2),))) == u[1]
    # ListTensor -- nested
    l0 = ListTensor(z, u[1], z)
    l1 = ListTensor(z, z, u[2])
    l2 = ListTensor(u[0], z, z)
    ll = ListTensor(l0, l1, l2)
    assert Indexed(ll, MultiIndex((FixedIndex(1), FixedIndex(2)))) == u[2]
    assert Indexed(ll, MultiIndex((FixedIndex(2), i))) == l2[i]
    # ComponentTensor + ListTensor
    c = ComponentTensor(
        Indexed(ll, MultiIndex((i, j))),
        MultiIndex((j, i))
    )
    assert Indexed(c, MultiIndex((FixedIndex(1), FixedIndex(2)))) == l2[1]
```

ksagiyam (approved):

test/test_simplify.py
```diff
@@ -193,3 +195,76 @@ def test_nested_indexed(self):
...
+    C = Coefficient(space)
+
+    r = len(shape)
+    kk = indices(r)
+
+    # Untangle as_tensor(C[kk], kk) -> C
+    B = as_tensor(Indexed(C, MultiIndex(kk)), kk)
+    assert B is C
+
+    # Untangle as_tensor(C[kk], jj)[ii] -> C[kk]
```
pbrubeck, line 234:
```suggestion
    # Untangle as_tensor(C[kk], jj)[ii] -> C[ll]
```

PR #334, opened 2024-12-18
Fix `is_cellwise_constant(ReferenceGrad(x))`

pbrubeck:
Adding a special case to detect that the reference gradient of a P1 field is cellwise constant.

PR #339, opened 2025-01-18
Remove component tensors

pbrubeck:
Consider the bilinear form
`a(v, u) = inner(u, v)*dx + inner(div(u), div(v))*dx`
where `u` is a (doubly) Piola mapped tensor in H(div, S) discretized with linear Johnson-Mercier macroelements.

Here we introduce two enhacements targeted to optimize code generation for `a(v, u)`:
1. We simplify `RefGrad(RefGrad(x)) -> 0` on affine simplex meshes. These zero terms would arise from the product rule on the doubly Piola mapped tensor.
1. We eagerly remove `ComponentTensors`, to make the UFL easier to traverse by the Form compiler. 

Since a `Form` always maps to a scalar, it is always possible to remove `ComponentTensor` nodes by pushing `Indexed` inside expressions and simplifying `Indexed(ComponentTensor)` with the approapiate index replacements.

This PR makes more agressive simplifications within the `Indexed` constructor, so it's pushed inside sums, and adds a multifunction to handle index replacements. The removal of `ComponentTensors` is applied as a final preprocessing step in `compute_form_data`.

### DIFF
--- a/ufl/algorithms/apply_derivatives.py
+++ b/ufl/algorithms/apply_derivatives.py
@@ -18,7 +18,6 @@ from ufl.argument import BaseArgument
 from ufl.checks import is_cellwise_constant
 from ufl.classes import (
     Coefficient,
-    ComponentTensor,
     Conj,
     ConstantValue,
     ExprList,
@@ -219,28 +218,12 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Indexing and component handling
 
-    def indexed(self, o, Ap, ii):  # TODO: (Partially) duplicated in nesting rules
+    def indexed(self, o, Ap, ii):
         """Differentiate an indexed."""
         # Propagate zeros
         if isinstance(Ap, Zero):
             return self.independent_operator(o)
 
-        # Untangle as_tensor(C[kk], jj)[ii] -> C[ll] to simplify
-        # resulting expression
-        if isinstance(Ap, ComponentTensor):
-            B, jj = Ap.ufl_operands
-            if isinstance(B, Indexed):
-                C, kk = B.ufl_operands
-                kk = list(kk)
-                if all(j in kk for j in jj):
-                    rep = dict(zip(jj, ii))
-                    Cind = [rep.get(k, k) for k in kk]
-                    expr = Indexed(C, MultiIndex(tuple(Cind)))
-                    assert expr.ufl_free_indices == o.ufl_free_indices
-                    assert expr.ufl_shape == o.ufl_shape
-                    return expr
-
-        # Otherwise a more generic approach
         r = len(Ap.ufl_shape) - len(ii)
         if r:
             kk = indices(r)
@@ -1450,29 +1433,12 @@ class DerivativeRuleDispatcher(MultiFunction):
             o_[3],
         )
 
-    def indexed(self, o, Ap, ii):  # TODO: (Partially) duplicated in generic rules
+    def indexed(self, o, Ap, ii):
         """Apply to an indexed."""
         # Reuse if untouched
         if Ap is o.ufl_operands[0]:
             return o
 
-        # Untangle as_tensor(C[kk], jj)[ii] -> C[ll] to simplify
-        # resulting expression
-        if isinstance(Ap, ComponentTensor):
-            B, jj = Ap.ufl_operands
-            if isinstance(B, Indexed):
-                C, kk = B.ufl_operands
-
-                kk = list(kk)
-                if all(j in kk for j in jj):
-                    rep = dict(zip(jj, ii))
-                    Cind = [rep.get(k, k) for k in kk]
-                    expr = Indexed(C, MultiIndex(tuple(Cind)))
-                    assert expr.ufl_free_indices == o.ufl_free_indices
-                    assert expr.ufl_shape == o.ufl_shape
-                    return expr
-
-        # Otherwise a more generic approach
         r = len(Ap.ufl_shape) - len(ii)
         if r:
             kk = indices(r)
--- a/ufl/algorithms/apply_derivatives.py
+++ b/ufl/algorithms/apply_derivatives.py
@@ -12,7 +12,6 @@ from math import pi
 
 from ufl.action import Action
 from ufl.algorithms.analysis import extract_arguments
-from ufl.algorithms.estimate_degrees import SumDegreeEstimator
 from ufl.algorithms.map_integrands import map_integrand_dags
 from ufl.algorithms.replace_derivative_nodes import replace_derivative_nodes
 from ufl.argument import BaseArgument
@@ -563,14 +562,6 @@ class GradRuleset(GenericDerivativeRuleset):
         """Initialise."""
         GenericDerivativeRuleset.__init__(self, var_shape=(geometric_dimension,))
         self._Id = Identity(geometric_dimension)
-        self.degree_estimator = SumDegreeEstimator(1, {})
-
-    def is_cellwise_constant(self, o):
-        """More precise checks for cellwise constants."""
-        if is_cellwise_constant(o):
-            return True
-        degree = map_expr_dag(self.degree_estimator, o)
-        return degree == 0
 
     # --- Specialized rules for geometric quantities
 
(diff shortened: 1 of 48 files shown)


ufl/algorithms/remove_component_tensors.py
```diff
@@ -0,0 +1,117 @@
...
+        if isinstance(o1, ComponentTensor):
+            # Simplify Indexed ComponentTensor
+            o2, i2 = o1.ufl_operands
+            # Replace inner indices first
+            v = map_expr_dag(self, o2)
+            # Replace outer indices
+            assert len(i2) == len(i1)
+            fimap = dict(zip(i2, i1))
+            rule = IndexReplacer(fimap)
+            return map_expr_dag(rule, v)
```
pbrubeck, line 94: @ksagiyam we need to be more general here. For instance the simplification `Indexed(LinearOp(ComponentTensor(x))) -> LinearOp(Indexed(ComponentTensor(x)))` can be made for `LinearOp in {Sum, IndexSum}`. 

This PR expands `Indexed(Sum)` and `Indexed(IndexSum)` at instantiation.
But maybe we only want to do that here. I think we should adopt as a convention that the __new__ constructor should only attempt to simplify, but never to expand.

pbrubeck: Also should these rules be merged into the `IndexExpander` https://github.com/firedrakeproject/ufl/blob/master/ufl/algorithms/expand_indices.py?

ufl/algorithms/remove_component_tensors.py
```diff
@@ -0,0 +1,117 @@
+"""Remove component tensors.
+
+This module contains classes and functions to remove component tensors.
+"""
+# Copyright (C) 2008-2016 Martin Sandve Alnæs
```
jorgensd, line 5: Update year and copyright notice?

ufl/algorithms/remove_component_tensors.py
```diff
@@ -0,0 +1,117 @@
...
+        """Simplify Indexed."""
+        o1, i1 = o.ufl_operands
+        if isinstance(o1, ComponentTensor):
+            # Simplify Indexed ComponentTensor
+            o2, i2 = o1.ufl_operands
+            # Replace inner indices first
+            v = map_expr_dag(self, o2)
+            # Replace outer indices
+            assert len(i2) == len(i1)
+            fimap = dict(zip(i2, i1))
```
jorgensd, line 92:
```suggestion
            # Replace outer indices
            # NOTE: Replace with `fimap = dict(zip(i2, i1, strict=True))` when
            # Pyhton>=3.10
            assert len(i2) == len(i1)
            fimap = dict(zip(i2, i1))
```

ufl/algorithms/remove_component_tensors.py
```diff
@@ -0,0 +1,117 @@
...
+            # Reuse if untouched
+            return o
+        return o._ufl_expr_reconstruct_(expr, i1)
+
+    reference_grad = _unary_operator
+    reference_value = _unary_operator
+
+
+def remove_component_tensors(o):
+    """Remove component tensors."""
```
jorgensd, line 107: Could we use `map_integrand_dags` here? @wence- pointed out the existence of this function to me last fall, and it cover way more logic than just form and Expr.

jorgensd:
In general this looks sensible to me. Some minor comments/suggestions only.

ufl/algorithms/apply_derivatives.py
```diff
@@ -562,6 +563,14 @@ def __init__(self, geometric_dimension):
         """Initialise."""
         GenericDerivativeRuleset.__init__(self, var_shape=(geometric_dimension,))
         self._Id = Identity(geometric_dimension)
+        self.degree_estimator = SumDegreeEstimator(1, {})
+
+    def is_cellwise_constant(self, o):
+        """More precise checks for cellwise constants."""
+        if is_cellwise_constant(o):
+            return True
+        degree = map_expr_dag(self.degree_estimator, o)
```
michalhabera, line 572: You've introduced here that `is_cellwise_constant` logic is based on degree estimation, which is the only place in the code. How does it work, since the degree estimation for geometric quantities calls `is_cellwise_constant` itself?   https://github.com/FEniCS/ufl/blob/1ab30c2de2a975b9fc957c6b356541bd15b35e15/ufl/algorithms/estimate_degrees.py#L46-L56

If there is no infinite loop, could this be used in the generic cell-wise constant check? Or is it that only `GradRuleset` does not cause infinite loop?

ufl/algorithms/apply_derivatives.py
```diff
@@ -653,9 +662,10 @@ def reference_value(self, o):
 
     def reference_grad(self, o):
         """Differentiate a reference_grad."""
+        if self.is_cellwise_constant(o):
```
michalhabera, line 665: Is this correct? Looks like one derivative is lost.

pbrubeck: Here we are simplifying `grad(o)` where `o = grad(f)`, if `o` is constant then we should return zero with the correc shape.

ufl/algorithms/apply_derivatives.py
pbrubeck: I first tried to add this logic to the generic `is_cellwise_constant` in https://github.com/FEniCS/ufl/pull/334, but simplifying `grad(cell_wise_constant)` on instantiation turns out to be problematic for nested `grad` expression simply because `grad(f)` expects f to have a domain and `Zero` does not have a domain, so  `grad(grad(grad(linear)))` would simplify to `grad(Zero)`.

ufl/algorithms/apply_derivatives.py
michalhabera: I see, thanks. I tried following simple UFL code

```python
import ufl
import ufl.algorithms
import ufl.algorithms.apply_algebra_lowering
import ufl.algorithms.apply_function_pullbacks
import ufl.utils.formatting


mesh_el = ufl.finiteelement.FiniteElement("P", ufl.Cell("triangle"), 1, (2, ), ufl.pullback.IdentityPullback(), ufl.sobolevspace.H1)
mesh = ufl.Mesh(mesh_el)

V_el = ufl.finiteelement.FiniteElement("JM", ufl.Cell("triangle"), 1, (2, 2), ufl.pullback.DoubleCovariantPiola(), ufl.sobolevspace.L2)
V = ufl.FunctionSpace(mesh, V_el)
u = ufl.Coefficient(V)
x = ufl.SpatialCoordinate(mesh)
a = ufl.div(u) 

a = ufl.algorithms.apply_algebra_lowering.apply_algebra_lowering(a)
a = ufl.algorithms.apply_derivatives.apply_derivatives(a)
a = ufl.algorithms.apply_function_pullbacks.apply_function_pullbacks(a)
a = ufl.algorithms.apply_derivatives.apply_derivatives(a)

print(ufl.utils.formatting.tree_format(a))
```
which mimics the order in `compute_form_data`. Yet, I do not see the `ReferenceGrad` applied to `JacobianInverse` in the tree. Geometric quantities makes it correctly outside of the `ReferenceGrad`. For non-affine cell `ReferenceGrad(JacobianInverse)` is present as expected. Do you have UFL snippet that shows the problem?

But lets assume that such node will exist and remain unsimplified even if it should. Would `ReferenceGradruleset.geometric_quatity` be more appropriate place for improved constness check?

michalhabera: What is always confusing is the order of preprocessing operations in `compute_form_data`. So my understanding is that for such expressions as `div(u)` where u is (some) Piola mapped you we essentially do
1. apply algebra lowering, `div(u) -> sum(grad(u)[i])`,
2. 1st derivate pass, `grad(u) -> grad(u)`,
3. function pullbacks, `grad(u) -> grad(piola_maps*reference_value(u))`,
4. 2nd derivative pass, `grad(piola_maps*u) -> piola_maps*reference_grad(reference_value(u))`,
5. 3rd derivative pass, `piola_maps*reference_grad(reference_value(u)) -> piola_maps*reference_grad(reference_value(u))`

In the code above I am seeing `GradRuleset.jacobian_inverse` hit constness of `JacobianInverse`, that makes it in step 4.

pbrubeck: In Firedrake we are getting lots of `grad(grad(x[j]))`. This could be because we do not include `JacobianInverse` in `preserve_geometry_types` https://github.com/firedrakeproject/firedrake/blob/96501362c73e17e28131294ccc7ee611b9b64e64/tsfc/ufl_utils.py#L38

pbrubeck:
I have removed the degree estimation, and handled `RefGrad(RefGrad(x)) -> 0` as a special case. @michalhabera could you review again?

PR #340, opened 2025-01-20
Simplify conditional

pbrubeck:
#Fixes https://github.com/firedrakeproject/firedrake/issues/3984

### DIFF
--- a/test/test_check_arities.py
+++ b/test/test_check_arities.py
@@ -95,14 +95,19 @@ def test_zero_simplify_arity():
     v = TestFunction(V)
     u = Coefficient(V)
 
+    nonzero = 1
+    with pytest.raises(ArityMismatch):
+        F = inner(u, v + nonzero) * dx
+        compute_form_data(F)
+
     zero = as_tensor([0, u])[0]
     F = inner(u, v + zero) * dx
-    compute_form_data(F, complex_mode=False)
+    compute_form_data(F)
 
     zero = conditional(u < 0, 0, 0)
     F = inner(u, v + zero) * dx
-    compute_form_data(F, complex_mode=False)
+    compute_form_data(F)
 
     zero = conditional(u < 0, 0, conditional(u == 0, 0, 0))
     F = inner(u, v + zero) * dx
-    compute_form_data(F, complex_mode=False)
+    compute_form_data(F)
--- a/test/test_check_arities.py
+++ b/test/test_check_arities.py
@@ -89,6 +89,10 @@ def test_product_arity():
 
 
 def test_zero_simplify_arity():
+    """
+    Test that adding verious zero-like expressions to a form is simplified,
+    such that one can compute form data for the integral.
+    """
     cell = tetrahedron
     D = Mesh(FiniteElement("Lagrange", cell, 1, (3,), identity_pullback, H1))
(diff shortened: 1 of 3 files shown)


test/test_check_arities.py
```diff
@@ -84,3 +86,23 @@ def test_product_arity():
...
+    F = inner(u, v + zero) * dx
+    compute_form_data(F, complex_mode=False)
+
+    zero = conditional(u < 0, 0, 0)
+    F = inner(u, v + zero) * dx
+    compute_form_data(F, complex_mode=False)
+
+    zero = conditional(u < 0, 0, conditional(u == 0, 0, 0))
+    F = inner(u, v + zero) * dx
+    compute_form_data(F, complex_mode=False)
```
jorgensd, line 108: Should these checks test some correctness, rather than checking that the code doesn’t throw and error at running?

pbrubeck: The function we are testing does not return anything, just raises the exception. Seems like this is the best we can do.

I also fixed some errors generated while raising the exception, and added a test.

test/test_check_arities.py
```diff
@@ -84,3 +86,28 @@ def test_product_arity():
...
+    F = inner(u, v + zero) * dx
+    compute_form_data(F)
+
+    zero = conditional(u < 0, 0, 0)
+    F = inner(u, v + zero) * dx
+    compute_form_data(F)
+
+    zero = conditional(u < 0, 0, conditional(u == 0, 0, 0))
+    F = inner(u, v + zero) * dx
+    compute_form_data(F)
```
jorgensd, line 113:
```suggestion

def test_zero_simplify_arity():
    """
    Test that adding verious zero-like expressions to a form is simplified,
    such that one can compute form data for the integral.
    """
    cell = tetrahedron
    D = Mesh(FiniteElement("Lagrange", cell, 1, (3,), identity_pullback, H1))
    V = FunctionSpace(D, FiniteElement("Lagrange", cell, 2, (), identity_pullback, H1))
    v = TestFunction(V)
    u = Coefficient(V)

    nonzero = 1
    with pytest.raises(ArityMismatch):
        F = inner(u, v + nonzero) * dx
        compute_form_data(F)
    z = Coefficient(V)

    # Add a Zero-component (rank-0) of a tensor to a rank-1 tensor
    zero = as_tensor([0, z])[0]
    F = inner(u, v + zero) * dx
    fd = compute_form_data(F)
    assert fd.num_coefficients == 1

    # Add a conditional that should have been simplified to zero (rank-0)
    # to a rank-1 tensor
    zero = conditional(z < 0, 0, 0)
    F = inner(u, v + zero) * dx
    fd = compute_form_data(F)
    assert fd.num_coefficients == 1

    # Check that nested zero conditionals are simplifed to zero (rank-0)
    # and can be added to a rank-1 tensor
    zero = conditional(z < 0, 0, conditional(z == 0, 0, 0))
    F = inner(u, v + zero) * dx
    fd = compute_form_data(F)
    assert fd.num_coefficients == 1
```
This is more along what I had in mind. Documenting each use-case and explaining what is expected to happen. I would like the extra documentation as someone might stumble upon this in many years in the future, and then some extra context often helps, or the behavior of `compute_form_data` might change, such that it can split forms.

jorgensd (approved):

pbrubeck:
Thanks @jorgensd

PR #341, opened 2025-01-27
Ksagiyam/add coefficient split

ksagiyam:


JHopeCollins:
Will this DTRT for `Cofunctions` too? They don't inherit from `Expr` so won't be caught by that single dispatch implementation.

ufl/corealg/dag_traverser.py
```diff
@@ -71,18 +71,18 @@ def process(self, o: Expr, *args) -> Expr:
         """
         raise AssertionError(f"Rule not set for {type(o)}")
 
-    def reuse_if_untouched(self, o: Expr) -> Expr:
+    def reuse_if_untouched(self, o: Expr, *args) -> Expr:
```
dham, line 74: let's do **kwargs instead.

ksagiyam: Done.

pbrubeck:
> Will this DTRT for `Cofunctions` too? They don't inherit from `Expr` so won't be caught by that single dispatch implementation.

I believe this is not an issue here, as this `DAGTraverser` is only meant to be applied to `Forms`

PR #348, opened 2025-02-19
Fix BaseFormOperator.ufl_function_space

pbrubeck:


dham (commented):
Looks good but needs tests. I think this should be easy to do using Interpolate as the BaseFormOperator in the tests.

pbrubeck:
> Looks good but needs tests. I think this should be easy to do using Interpolate as the BaseFormOperator in the tests.

The failing case would need to construct an `Interpolate(X, Cofunction)`. Is this a resonable object @dham?

dham:
> > Looks good but needs tests. I think this should be easy to do using Interpolate as the BaseFormOperator in the tests.
> 
> The failing case would need to construct an `Interpolate(X, Cofunction)`. Is this a resonable object @dham?

I think I can't see a failing test (the tsfc thing looks to be different). However `Interpolate(X, Cofunction)` is completely reasonable. It does the adjoint interpolation of `Cofunction` and applies it to `X`. If `X` is an `Argument`, the application does nothing so you just get the interpolated `Cofunction` object back.

pbrubeck:
I meant the _previously_ failing case

connorjward:
This is now one of the final blocks for pip install Firedrake so it would be amazing if this could get merged soon.

mscroggs:
@dham Happy for you to hit merge on this if you're happy that the testing is now sufficient

dham (approved):

PR #350, opened 2025-02-22
Extract lhs and rhs from `MixedFunctionSpace`

jorgensd:
Based of @cdaversin branch.
Corrects some bugs and adds a test in UFL for this behavior.


ufl/algorithms/formtransformations.py
```diff
@@ -362,7 +359,18 @@ def compute_form_lhs(form):
         a = u*v*dx + f*v*dx
         a = lhs(a) -> u*v*dx
     """
-    return compute_form_with_arity(form, 2)
+    parts = tuple(sorted(set(part for a in form.arguments() if (part := a.part()) is not None)))
+    # If Arguments are stemming from a MixedFucntionSpace, we have to compute this per block
```
jorgensd, line 363:
```suggestion
    # If Arguments are stemming from a MixedFunctionSpace, we have to compute this per block
```

ufl/algorithms/formsplitter.py
```diff
@@ -24,8 +24,8 @@
 class FormSplitter(MultiFunction):
     """Form splitter."""
 
-    def split(self, form, ix, iy=0):
-        """Split."""
+    def split(self, form, ix, iy=None):
+        """Split form based on the arugment part/number."""
```
jorgensd, line 28:
```suggestion
        """Split form based on the argument part/number."""
```

ufl/algorithms/formsplitter.py
```diff
@@ -147,16 +155,16 @@ def extract_blocks(form, i: Optional[int] = None, j: Optional[None] = None):
         if arity > 1:
             for pj in range(num_parts):
                 f = fs.split(form, pi, pj)
-                if f.empty():
+                # Ignore empty forms and linear forms
+                if f.empty() or len(f.arguments()) != 2:
```
cdaversin, line 159: Do we ever get `arity > 1` and  `len(f.arguments()) != 2` ? Should we (at least) raise a Warning in that case, just to make sure we don't have a term in the form that has not been taken into account (might be due to user mistake, but as it is the user won't know)

ufl/algorithms/formtransformations.py
```diff
@@ -362,7 +359,18 @@ def compute_form_lhs(form):
...
+    if parts == ():
+        return compute_form_with_arity(form, 2)
+
+    form_blocks = extract_blocks(form, arity=2)
+    lhs = 0
+    for bi in form_blocks:
+        for bj in bi:
+            if bj is not None:
+                lhs += compute_form_with_arity(bj, 2)
+    return lhs
```
cdaversin, line 373: In the test above, we want to extract the blocks of the left and right hand sides of our form. What it does : 
(1) `lhs(F)` extracts the blocks and sum the corresponding subforms : `lhs += compute_form_with_arity(bj, 2)` 
(2) we do `a_blocked = extract_blocks(a)` where `a=lhs(F)`, meaning re-doing what the block extraction already implemented in `lhs()`
Just an idea: Would it be worth having functions `lhs_blocks()` and `rhs_block()`, or maybe some option to `lhs()` and `rhs()` to return the list of blocks and not the sum ?

cdaversin (commented):
This looks good to me! Only a few minor suggestions (see my comments)

ufl/algorithms/formsplitter.py
```diff
@@ -147,16 +155,16 @@ def extract_blocks(form, i: Optional[int] = None, j: Optional[None] = None):
         if arity > 1:
             for pj in range(num_parts):
                 f = fs.split(form, pi, pj)
-                if f.empty():
+                # Ignore empty forms and linear forms
```
cdaversin, line 158:
```suggestion
                # Ignore empty forms and rank 0 or 1 forms
```

ufl/algorithms/formsplitter.py
jorgensd: Yes, formsplitter can return a linear or functional at this point.

ufl/algorithms/formtransformations.py
jorgensd: I don't think so, as on blocked structures (lists or nested lists), UFL cant perform any symbolical magic. Therefore, we want the final step of the algorithm before calling anything in DOLFINx to be `ufl.extract_blocks`.

cdaversin (approved):

PR #351, opened 2025-02-22
Add `ufl.action` for MixedFunctionSpace

jorgensd:
Extending #350, such that we can use `ufl.action` on forms from `MixedFunctionSpace`.


ufl/algorithms/formtransformations.py
```diff
@@ -1,6 +1,6 @@
 """Utilities for transforming complete Forms into new related Forms."""
 
-# Copyright (C) 2008-2016 Martin Sandve Alnæs
+# Copyright (C) 2008-2015 Martin Sandve Alnæs
```
jorgensd, line 3:
```suggestion
# Copyright (C) 2008-2025 Martin Sandve Alnæs
```

ufl/algorithms/formtransformations.py
```diff
@@ -416,7 +417,26 @@ def compute_form_action(form, coefficient):
 
     parts = [arg.part() for arg in arguments]
     if set(parts) - {None}:
-        raise ValueError("compute_form_action cannot handle parts.")
+        # We assume that for MixedFunctionSpace that the max arity can be two
+        highest_arity_form = compute_form_lhs(form)
+        if highest_arity_form == 0 or highest_arity_form.empty():
+            highest_arity_form = compute_form_rhs(form)
+        if highest_arity_form == 0 or highest_arity_form.empty():
```
cdaversin, line 424: Shouldn't this `if` be within the previous one to avoid testing the same thing twice ? 
```
if highest_arity_form == 0 or highest_arity_form.empty():
    highest_arity_form = compute_form_rhs(form)
    if highest_arity_form == 0 or highest_arity_form.empty():
        raise ValueError("No arguments to replace in form.")
```

cdaversin (commented):
Looks good to me :)
Just out of curiosity : In the standard case (not MixedFunctionSpace), we pick the last argument : `u = arguments[-1]`.
For a `MixedFunctionSpace`, we choose the argument `a` which has the maximal `a.number()`. Why don't we just choose `arguments[-1]` from the highest arity form ?

ufl/algorithms/formtransformations.py
```diff
@@ -416,7 +417,26 @@ def compute_form_action(form, coefficient):
...
+            raise ValueError("No arguments to replace in form.")
```
jorgensd, line 425:
```suggestion
            if highest_arity_form == 0 or highest_arity_form.empty():
                raise ValueError("No arguments to replace in form.")
```

ufl/algorithms/formtransformations.py
jorgensd: Agreed

jorgensd:
> Looks good to me :) Just out of curiosity : In the standard case (not MixedFunctionSpace), we pick the last argument : `u = arguments[-1]`. For a `MixedFunctionSpace`, we choose the argument `a` which has the maximal `a.number()`. Why don't we just choose `arguments[-1]` from the highest arity form ?

Multiple arguments (one per part of the mixed space)

ufl/algorithms/formsplitter.py
```diff
@@ -25,7 +25,7 @@ class FormSplitter(MultiFunction):
     """Form splitter."""
 
     def split(self, form, ix, iy=None):
-        """Split form based on the argument part/number."""
+        """Split form based on the arugment part/number."""
```
jorgensd, line 28:
```suggestion
        """Split form based on the argument part/number."""
```

PR #356, opened 2025-03-02
Some `mypy` checks

schnellerhase:
Work towards https://github.com/FEniCS/ufl/issues/211


ufl/cell.py
```diff
@@ -340,7 +340,7 @@ def reconstruct(self, **kwargs: typing.Any) -> Cell:
         return Cell(self._cellname)
 
 
-class TensorProductCell(AbstractCell):
+class TensorProductCell(Cell):
```
jorgensd, line 343: Was this change deliberate? I guess we should talk to our firedrake friends (@dham) if this is  something they want.

ufl/finiteelement.py
```diff
@@ -53,7 +53,7 @@ def __hash__(self) -> int:
         """Return a hash."""
 
     @_abc.abstractmethod
-    def __eq__(self, other: AbstractFiniteElement) -> bool:
```
jorgensd, line 56: This confuses me a bit. Do we want a more general input to be supported for equality checks than something that inherits `AbstractFiniteElement`? Similar comment for `__ne__`.

schnellerhase: `abc.ABC` defines it with signature `other: object`, which makes it necessary in derived class as well.

ufl/cell.py
schnellerhase: Yes, fixes 
```
ufl/cell.py:441: error: Incompatible return value type (got "TensorProductCell", expected "Cell")  [return-value]
```
Not entirely sure if it is the right fix to it though.

ufl/finiteelement.py
jorgensd: I guess that makes sense, as one in theory could check that an abstractfiniteelement is equal to something of a different class.

ufl/cell.py
mscroggs: Perhaps the return type of `reconstruct` should be `AbstractCell` rather than `Cell`? This this can go back to inheriting from `AbstractCell`?

garth-wells (approved):

schnellerhase:
Verified for firedrake at https://github.com/firedrakeproject/firedrake/pull/4343, thanks @connorjward.

PR #360, opened 2025-03-18
Fix replace_derivative_nodes

pbrubeck:
Fixes https://github.com/firedrakeproject/firedrake/issues/4120

This PR implements coefficient derivatives on more generic ufl types such as `Cofunctions`.

Inspired from `DerivativeRuleDispatcher`, the right thing would be https://github.com/FEniCS/ufl/blob/069f07810dc13fa69ecda6f7ad6de41bc287f367/ufl/algorithms/apply_derivatives.py#L1454


michalhabera (approved):
This looks correct. But the function has very confusing name. I'd rather call it `replace_coefficient_derivative_directions`...

PR #361, opened 2025-03-21
Fix missing domain kwarg in pullback

pbrubeck:


michalhabera:
What is the reason for this change? None of the `FooPullback.apply()` have the domain kwarg. So this is "missing" in what sense? I'd expect all other pullbacks that extract the domain from the expression require the same change then.

pbrubeck:
> What is the reason for this change? None of the `FooPullback.apply()` have the domain kwarg. So this is "missing" in what sense? I'd expect all other pullbacks that extract the domain from the expression require the same change then.

What do you mean, all the Piola pullbacks do this, see here for example:
https://github.com/FEniCS/ufl/blob/82756545eb8d5d541df31ec6d01c7b9dec9227ac/ufl/pullback.py#L318

pbrubeck:
This pullback was neglected by a recent PR https://github.com/FEniCS/ufl/pull/303 that introduced this new kwarg in `apply()`.

michalhabera:
> This pullback was neglected by a recent PR #303 that introduced this new kwarg in `apply()`.

Ah, all good, I was looking at the UFL before that PR.

Issue #35, opened 2020-07-21
DAG visitor overhaul


wence-:
The `MultiFunction` based `map_dag` visitors in UFL provide a caching (and hence efficient) API for implementing post-order visits of a UFL DAG. The core function is `map_expr_dags` (the relevant part is here https://github.com/FEniCS/ufl/blob/master/ufl/corealg/map_dag.py#L64).

Most of the transformations in the UFL pipeline are bottom up (post-order) and so are well-served by this infrastructure. Some, however, are pre-order (or at least partially pre-order). Examples include the restriction propagation (which hand-codes a post-order visit of a restricted subtree https://github.com/FEniCS/ufl/blob/master/ufl/algorithms/apply_restrictions.py#L36)

This can lead to unintended performance bugs where one, by necessity, hand-codes the DAG recursion to get a pre-order visit. This will blow up the perceived size of the DAG, the avoidance of which was one of the major motiviating factors for the `ReuseTransformer` -> `MultiFunction + map_expr_dag` transition.

Unfortunately, providing a caching pre-order visit requires that the visitor object sees the cache (the post-order visit cache can be managed completely from the outside). In the latter case the `MultiFunction` does not implement its own recursion and instead is used for type-based method dispatch on the first argument.

In tsfc, this issue is solved by having a more general visitor implementation (https://github.com/firedrakeproject/tsfc/blob/master/gem/node.py#L187). One writes `singledispatch` recursive functions that take the cache-manager as a parameter. Recursion is done by `__call__`ing this parameter on children as necessary.

Here's an example from tsfc using this pattern to manage `abs`-simplification (from https://github.com/firedrakeproject/tsfc/blob/master/tsfc/ufl_utils.py#L250)

```python
@singledispatch
def _simplify_abs(o, self, in_abs):
    """Single-dispatch function to simplify absolute values.

    :arg o: UFL node
    :arg self: Callback handler for recursion
    :arg in_abs: Is ``o`` inside an absolute value?

    When ``in_abs`` we must return a non-negative value, potentially
    by wrapping the returned node with ``Abs``.
    """
    raise AssertionError("UFL node expected, not %s" % type(o))

@_simplify_abs.register(Expr)
def _simplify_abs_expr(o, self, in_abs):
    # General case, only wrap the outer expression (if necessary)
    operands = [self(op, False) for op in o.ufl_operands]
    result = ufl_reuse_if_untouched(o, *operands)
    if in_abs:
        result = Abs(result)
    return result

...

[...]

Issue #176, opened 2023-06-29
Use `typing.Self` once minimum Python version is 3.11


mscroggs:
After 04 Oct 2026 (Python 3.10 end of life), we should start using `typing.Self` to annotate function input types (ie undo https://github.com/FEniCS/ufl/commit/86a2140aa0ea3c65e5d96de1ad8f2bfc06347a60).

PR #365, opened 2025-03-26
add DAGTraverser class and remove MultiFunctions in apply_derivatives()

ksagiyam:
Attempt to implement what @wence- described in [this issue](https://github.com/FEniCS/ufl/issues/35).

We basically would like to introduce classes (`DAGTraverser`s) that define node processing (using `singledispatchmethod`) and hold caches.

We considered two DAG traversal approaches (for post-order and mostly post-order traversals).

**Approach 1. Two-step approach.**

Collect nodes first post-order and then process them in order.

This is the current UFL approach (`cutoff_unique_post_traversal` + `MultiFunction` wrapped in `map_expr_dags`).

This "bottom-up" approach is optimal, but when we process a node, we no longer know the local DAG structure around that node (e.g., we do not know its parent). So if we want to pass down some state from the parent, we need to (indirectly) specify that that parent is a "cutoff" node (so that we do not collect the child nodes in the first step), and perform a sub-DAG traversal under that parent; see [RestrictionPropagator.restricted(self, o)](https://github.com/FEniCS/ufl/blob/b002eb0e815f74c8ed9ca3b53aa77e2d2486f062/ufl/algorithms/apply_restrictions.py#L40). In the current implementation, if the number of arguments to the method is two (`self` and `o` in the above case), the corresponding node type is regarded as a cutoff node type. `MultiFunction` constructor currently identifies cutoff node types by `inspect`ing the signature of each method. If we wanted to do a similar thing with `singledispatchmethod`, we could subclass `singledispatchmethod` as David suggested, but we found that we would end up overwriting `singledispatchmethod.register()` method relying on the current implementation of `singledispatchmethod`, which would be a drawback.

**Approach 2. Monolithic approach.**

Use recursion with caching (each node is processed once and only once) described in [this issue](https://github.com/FEniCS/ufl/issues/35).

I observed about a few % - 20% overhead after rewriting `apply_derivatives.py` (please see below) presumably due to recursion, but no special consideration of cutoff is required. This approach is claimed to be more robust.

##

In this PR we take Approach2, and replace `MultiFunction` s with `DAGTraverser`s in `apply_derivatives.py`. We should be able to incrementally/systematically remove all `MultiFunction`s in the future.

Performance checks:

**`holzapfel_ogden.py`** ([holzapfel_ogden](https://gist.github.com/FabioLuporini/a74baa84b64fd1c08f5eb9cc45c8ea7e)):
```
import time

from ufl import (
    Coefficient,
    Constant,
    FunctionSpace,
    Identity,
    Mesh,
    TestFunction,
    derivative,
    det,
    diff,
    dot,
    dx,
    exp,
    grad,
    inner,
    ln,
    tetrahedron,
    tr,
    variable,
)
from ufl.algorithms import compute_form_data


mesh = Mesh(FiniteElement("Lagrange", tetrahedron, 1, (3,), identity_pullback, H1))
lamda = Constant(mesh)
a = Constant(mesh)
b = Constant(mesh)
a_s = Constant(mesh)
b_s = Constant(mesh)
a_f = Constant(mesh)
b_f = Constant(mesh)
a_fs = Constant(mesh)
b_fs = Constant(mesh)
e_s = Constant(mesh, shape=(3,))
e_f = Constant(mesh, shape=(3,))

def isochoric(F):
    C = F.T*F
    I_1 = tr(C)
    I4_f = dot(e_f, C*e_f)
    I4_s = dot(e_s, C*e_s)
    I8_fs = dot(e_f, C*e_s)

    def cutoff(x):
        return 1.0/(1.0 + exp(-(x - 1.0)*30.0))

    def scaled_exp(a0, a1, argument):
        return a0/(2.0*a1)*(exp(b*argument) - 1)

    def scaled_exp(a0, a1, argument):
        return a0/(2.0*a1)*(exp(b*argument) - 1)

    E_1 = scaled_exp(a, b, I_1 - 3.)
    E_f = cutoff(I4_f)*scaled_exp(a_f, b_f, (I4_f - 1.)**2)
    E_s = cutoff(I4_s)*scaled_exp(a_s, b_s, (I4_s - 1.)**2)
    E_3 = scaled_exp(a_fs, b_fs, I8_fs**2)
    E = E_1 + E_f + E_s + E_3
    return E

elem = FiniteElement("Lagrange", tetrahedron, 1, (3,), identity_pullback, H1)
V = FunctionSpace(mesh, elem)
u = Coefficient(V)
v = TestFunction(V)
I = Identity(mesh.ufl_cell().topological_dimension())
F = grad(u) + I
F = variable(F)
J = det(F)
Fbar = J**(-1.0/3.0)*F
E_volumetric = lamda*0.5*ln(J)**2
psi = isochoric(Fbar) + E_volumetric
P = diff(psi, F)
F = inner(P, grad(v))*dx
a = derivative(F, u)

ntest = 2
time_sum = 0
for _ in range(ntest):
    start = time.time()
    fd = compute_form_data(
        a,
        do_apply_function_pullbacks=True,
        do_apply_default_restrictions=True,
        do_apply_geometry_lowering=True,
        do_apply_restrictions=True,
        complex_mode=False,
    )
    end = time.time()
    time_sum += end - start
print("average time required: ", time_sum / ntest)
```
`main`:
average time required:  0.27037227153778076

`this PR`:
average time required:  0.27785348892211914 (+2.8%)

With https://github.com/FEniCS/ufl/pull/339 (<- performance regression) reverted:

`main`:
average time required:  0.04610729217529297

`this PR`:
average time required:  0.055258870124816895 (+20%)

**`wence_test`** (https://github.com/FEniCS/ufl/pull/69):
```
import time

from ufl import (
    Coefficient,
    Constant,
    FacetNormal,
    FunctionSpace,
    Identity,
    Mesh,
    avg,
    derivative,
    det,
    diff,
    dS,
    grad,
    hexahedron,
    inner,
    ln,
    outer,
    variable,
)
from ufl.algorithms import compute_form_data
from ufl.pullback import identity_pullback, contravariant_piola
from ufl.sobolevspace import H1, HDiv

cell = hexahedron
mesh = Mesh(FiniteElement("Q", cell, 1, (3,), identity_pullback, H1))
V = FunctionSpace(mesh, FiniteElement("NCF", cell, 2, (3,), contravariant_piola, HDiv))
u = Coefficient(V)
mu = Constant(mesh)
psi = lambda F: (mu/2) * (inner(F, F) - u.ufl_shape[0]) - mu*ln(det(F))
flux = lambda F: diff(psi(F), F)
n = FacetNormal(mesh)
uxn = outer(u('+'), n('+')) + outer(u('-'), n('-'))
eye = Identity(u.ufl_shape[0])
grad_u = grad(u)
F_ = variable(grad_u + eye)
Fm = variable(grad_u('-') + eye)
Fp = variable(grad_u('+') + eye)
avg_flux = avg(flux(F_))
U = inner(avg_flux, uxn)*dS
a = derivative(derivative(U, u), u)

ntest = 2
time_sum = 0
for _ in range(ntest):
    start = time.time()
    fd = compute_form_data(
        a,
        do_apply_function_pullbacks=True,
        do_apply_default_restrictions=True,
        do_apply_geometry_lowering=True,
        do_apply_restrictions=True,
        complex_mode=False,
    )
    end = time.time()
    time_sum += end - start
print("average time required: ", time_sum / ntest)
```
`main`:
average time required:  0.23360061645507812

`this PR`:
average time required:  0.2485201358795166 (+6.4%)

With https://github.com/FEniCS/ufl/pull/339 reverted:

`main`:
average time required:  0.060405850410461426

`this PR`:
average time required:  0.06431174278259277 (+6.5%)

**Firedrake CI**
https://github.com/firedrakeproject/firedrake/pull/4145 (no notable performance regression).

### DIFF
--- a/ufl/algorithms/apply_derivatives.py
+++ b/ufl/algorithms/apply_derivatives.py
@@ -7,26 +7,36 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
 import warnings
-from collections import defaultdict
+from functools import singledispatchmethod
 from math import pi
+from typing import Union
 
 import numpy as np
 
 from ufl.action import Action
 from ufl.algorithms.analysis import extract_arguments, extract_coefficients
-from ufl.algorithms.map_integrands import map_integrand_dags
+from ufl.algorithms.map_integrands import map_integrands
 from ufl.algorithms.replace_derivative_nodes import replace_derivative_nodes
-from ufl.argument import BaseArgument
+from ufl.argument import Argument, BaseArgument, Coargument
+from ufl.averaging import CellAvg, FacetAvg
 from ufl.checks import is_cellwise_constant
 from ufl.classes import (
+    Abs,
+    CellCoordinate,
     Coefficient,
+    Cofunction,
+    ComponentTensor,
     Conj,
+    Constant,
     ConstantValue,
+    Division,
+    Expr,
     ExprList,
     ExprMapping,
     FacetNormal,
     FloatValue,
     FormArgument,
+    GeometricQuantity,
     Grad,
     Identity,
     Imag,
@@ -35,7 +45,9 @@ from ufl.classes import (
     Jacobian,
     JacobianDeterminant,
     JacobianInverse,
+    Label,
     ListTensor,
+    Power,
     Product,
     Real,
     ReferenceGrad,
@@ -45,21 +57,50 @@ from ufl.classes import (
     Variable,
     Zero,
 )
+from ufl.conditional import BinaryCondition, Conditional, NotCondition
 from ufl.constantvalue import is_true_ufl_scalar, is_ufl_scalar
 from ufl.core.base_form_operator import BaseFormOperator
 from ufl.core.expr import ufl_err_str
+from ufl.core.external_operator import ExternalOperator
+from ufl.core.interpolate import Interpolate
 from ufl.core.multiindex import FixedIndex, MultiIndex, indices
 from ufl.core.terminal import Terminal
-from ufl.corealg.map_dag import map_expr_dag
-from ufl.corealg.multifunction import MultiFunction
+from ufl.corealg.dag_traverser import DAGTraverser
 from ufl.differentiation import (
     BaseFormCoordinateDerivative,
     BaseFormOperatorDerivative,
+    CoefficientDerivative,
     CoordinateDerivative,
+    Derivative,
+    VariableDerivative,
 )
 from ufl.domain import MeshSequence, extract_unique_domain
 from ufl.form import BaseForm, Form, ZeroBaseForm
+from ufl.mathfunctions import (
+    Acos,
+    Asin,
+    Atan,
+    Atan2,
+    BesselI,
+    BesselJ,
+    BesselK,
+    BesselY,
+    Cos,
+    Cosh,
+    Erf,
+    Exp,
+    Ln,
+    MathFunction,
+    Sin,
+    Sinh,
+    Sqrt,
+    Tan,
+    Tanh,
+)
+from ufl.matrix import Matrix
 from ufl.operators import (
+    MaxValue,
+    MinValue,
     bessel_I,
     bessel_J,
     bessel_K,
@@ -77,6 +118,7 @@ from ufl.operators import (
     sqrt,
 )
 from ufl.pullback import CustomPullback, PhysicalPullback
+from ufl.restriction import Restricted
 from ufl.tensors import as_scalar, as_scalars, as_tensor, unit_indexed_tensor, unwrap_list_tensor
 
 # TODO: Add more rulesets?
@@ -107,22 +149,19 @@ def flatten_domain_element(domain, element):
     return flattened
 
 
-class GenericDerivativeRuleset(MultiFunction):
+class GenericDerivativeRuleset(DAGTraverser):
     """A generic derivative."""
 
-    def __init__(self, var_shape):
+    def __init__(
+        self,
+        var_shape: tuple,
+        compress: Union[bool, None] = True,
+        visited_cache: Union[dict[tuple, Expr], None] = None,
+        result_cache: Union[dict[Expr, Expr], None] = None,
+    ) -> None:
         """Initialise."""
-        MultiFunction.__init__(self)
+        super().__init__(compress=compress, visited_cache=visited_cache, result_cache=result_cache)
         self._var_shape = var_shape
-
-    # --- Error checking for missing handlers and unexpected types
-
-    def expr(self, o):
-        """Raise error."""
-        raise ValueError(
-            f"Missing differentiation handler for type {o._ufl_class_.__name__}. "
-            "Have you added a new type?"
-        )
 
     def unexpected(self, o):
         """Raise error about unexpected type."""
@@ -132,13 +171,6 @@ class GenericDerivativeRuleset(MultiFunction):
         """Raise error about overriding."""
         raise ValueError(
             f"Type {o._ufl_class_.__name__} must be overridden in specialized AD rule set."
-        )
-
-    def derivative(self, o):
-        """Raise error."""
-        raise ValueError(
-            f"Unhandled derivative type {o._ufl_class_.__name__}, "
-            "nested differentiation has failed."
         )
 
     # --- Some types just don't have any derivative, this is just to
@@ -152,9 +184,6 @@ class GenericDerivativeRuleset(MultiFunction):
         """
         return o
 
-    label = non_differentiable_terminal
-    multi_index = non_differentiable_terminal
-
     # --- Helper functions for creating zeros with the right shapes
 
     def independent_terminal(self, o):
@@ -165,28 +194,70 @@ class GenericDerivativeRuleset(MultiFunction):
         """A zero with correct shape and indices for operators independent of diff. variable."""
         return Zero(o.ufl_shape + self._var_shape, o.ufl_free_indices, o.ufl_index_dimensions)
 
+    # --- Error checking for missing handlers and unexpected types
+
+    @singledispatchmethod
+    def process(self, o: Expr) -> Expr:
+        """Process ``o``.
+
+        Args:
+            o: `Expr` to be processed.
+
+        Returns:
+            Processed object.
+
+        """
+        return super().process(o)
+
+    @process.register(Expr)
+    def _(self, o: Expr) -> Expr:
+        """Raise error."""
+        raise ValueError(
+            f"Missing differentiation handler for type {o._ufl_class_.__name__}. "
+            "Have you added a new type?"
+        )
+
+    @process.register(Derivative)
+    def _(self, o: Expr) -> Expr:
+        """Raise error."""
+        raise ValueError(
+            f"Unhandled derivative type {o._ufl_class_.__name__}, "
+            "nested differentiation has failed."
+        )
+
+    @process.register(Label)
+    @process.register(MultiIndex)
+    def _(self, o: Expr) -> Expr:
+        return self.non_differentiable_terminal(o)
+
     # --- All derivatives need to define grad and averaging
 
-    grad = override
-    cell_avg = override
-    facet_avg = override
+    @process.register(Grad)
+    @process.register(CellAvg)
+    @process.register(FacetAvg)
+    def _(self, o: Expr) -> Expr:
+        return self.override(o)
 
     # --- Default rules for terminals
 
     # Literals are by definition independent of any differentiation variable
-    constant_value = independent_terminal
-
+    @process.register(ConstantValue)
     # Constants are independent of any differentiation
-    constant = independent_terminal
+    @process.register(Constant)
+    def _(self, o: Expr) -> Expr:
+        return self.independent_terminal(o)
 
     # Zero may have free indices
-    zero = independent_operator
+    @process.register(Zero)
+    def _(self, o: Expr) -> Expr:
+        return self.independent_operator(o)
 
     # Rules for form arguments must be specified in specialized rule set
-    form_argument = override
-
+    @process.register(FormArgument)
     # Rules for geometric quantities must be specified in specialized rule set
-    geometric_quantity = override
+    @process.register(GeometricQuantity)
+    def _(self, o: Expr) -> Expr:
+        return self.override(o)
 
     # These types are currently assumed independent, but for non-affine domains
     # this no longer holds and we want to implement rules for them.
@@ -239,18 +310,21 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Default rules for operators
 
-    def variable(self, o, df, unused_l):
+    @process.register(Variable)
+    def _(self, o: Expr) -> Expr:
         """Differentiate a variable."""
-        return df
+        op, _ = o.ufl_operands
+        return self(op)
 
     # --- Indexing and component handling
 
-    def indexed(self, o, Ap, ii):
+    @process.register(Indexed)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, Ap: Expr, ii: Expr) -> Expr:
         """Differentiate an indexed."""
         # Propagate zeros
         if isinstance(Ap, Zero):
             return self.independent_operator(o)
-
         r = len(Ap.ufl_shape) - len(ii)
         if r:
             kk = indices(r)
@@ -260,11 +334,14 @@ class GenericDerivativeRuleset(MultiFunction):
             op = Indexed(Ap, ii)
         return op
 
-    def list_tensor(self, o, *dops):
+    @process.register(ListTensor)
+    def _(self, o: Expr) -> Expr:
         """Differentiate a list_tensor."""
-        return ListTensor(*dops)
+        return ListTensor(*(self(op) for op in o.ufl_operands))
 
-    def component_tensor(self, o, Ap, ii):
+    @process.register(ComponentTensor)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, Ap: Expr, ii: Expr) -> Expr:
         """Differentiate a component_tensor."""
         if isinstance(Ap, Zero):
             op = self.independent_operator(o)
@@ -275,15 +352,21 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Algebra operators
 
-    def index_sum(self, o, Ap, i):
+    @process.register(IndexSum)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, Ap: Expr, ii: Expr) -> Expr:
         """Differentiate an index_sum."""
-        return IndexSum(Ap, i)
+        return IndexSum(Ap, ii)
 
-    def sum(self, o, da, db):
+    @process.register(Sum)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, da: Expr, db: Expr) -> Expr:
         """Differentiate a sum."""
         return da + db
 
-    def product(self, o, da, db):
+    @process.register(Product)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, da: Expr, db: Expr) -> Expr:
         """Differentiate a product."""
         # Even though arguments to o are scalar, da and db may be
         # tensor valued
@@ -296,20 +379,19 @@ class GenericDerivativeRuleset(MultiFunction):
             s = as_tensor(s, ii)
         return s
 
-    def division(self, o, fp, gp):
+    @process.register(Division)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr, gp: Expr) -> Expr:
         """Differentiate a division."""
         f, g = o.ufl_operands
-
         if not is_ufl_scalar(f):
             raise ValueError("Not expecting nonscalar nominator")
         if not is_true_ufl_scalar(g):
             raise ValueError("Not expecting nonscalar denominator")
-
         # do_df = 1/g
         # do_dg = -h/g
         # op = do_df*fp + do_df*gp
         # op = (fp - o*gp) / g
-
         # Get o and gp as scalars, multiply, then wrap as a tensor
         # again
         so, oi = as_scalar(o)
@@ -318,23 +400,21 @@ class GenericDerivativeRuleset(MultiFunction):
         if oi or gi:
             o_gp = as_tensor(o_gp, oi + gi)
         op = (fp - o_gp) / g
-
         return op
 
-    def power(self, o, fp, gp):
+    @process.register(Power)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr, gp: Expr) -> Expr:
         """Differentiate a power."""
         f, g = o.ufl_operands
-
         if not is_true_ufl_scalar(f):
             raise ValueError("Expecting scalar expression f in f**g.")
         if not is_true_ufl_scalar(g):
             raise ValueError("Expecting scalar expression g in f**g.")
-
         # Derivation of the general case: o = f(x)**g(x)
         # do/df  = g * f**(g-1) = g / f * o
         # do/dg  = ln(f) * f**g = ln(f) * o
         # do/df * df + do/dg * dg = o * (g / f * df + ln(f) * dg)
-
         if isinstance(gp, Zero):
             # This probably produces better results for the common
             # case of f**constant
@@ -345,7 +425,6 @@ class GenericDerivativeRuleset(MultiFunction):
             op = f ** (g - 1) * (
                 g * fp + f * ln(f) * gp
             )  # This gives better accuracy in dolfin integration test
-
         # Example: d/dx[x**(x**3)]:
         # f = x
         # g = x**3
@@ -355,10 +434,11 @@ class GenericDerivativeRuleset(MultiFunction):
         #     = x**(x**3)   * (x**3/x + 3*x**2*ln(x))
         # op2 = f**(g-1) * (g*fp + f*ln(f)*gp)
         #     = x**(x**3-1) * (x**3 + x*3*x**2*ln(x))
-
         return op
 
-    def abs(self, o, df):
+    @process.register(Abs)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr) -> Expr:
         """Differentiate an abs."""
         (f,) = o.ufl_operands
         # return conditional(eq(f, 0), 0, Product(sign(f), df)) abs is
@@ -369,70 +449,96 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Complex algebra
 
-    def conj(self, o, df):
+    @process.register(Conj)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr) -> Expr:
         """Differentiate a conj."""
         return Conj(df)
 
-    def real(self, o, df):
+    @process.register(Real)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr) -> Expr:
         """Differentiate a real."""
         return Real(df)
 
-    def imag(self, o, df):
+    @process.register(Imag)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr) -> Expr:
         """Differentiate a imag."""
         return Imag(df)
 
     # --- Mathfunctions
 
-    def math_function(self, o, df):
+    @process.register(MathFunction)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr) -> Expr:
         """Differentiate a math_function."""
         # FIXME: Introduce a UserOperator type instead of this hack
         # and define user derivative() function properly
         if hasattr(o, "derivative"):
-            (f,) = o.ufl_operands
             return df * o.derivative()
-        raise ValueError("Unknown math function.")
+        else:
+            raise ValueError("Unknown math function.")
 
-    def sqrt(self, o, fp):
+    @process.register(Sqrt)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a sqrt."""
         return fp / (2 * o)
 
-    def exp(self, o, fp):
+    @process.register(Exp)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate an exp."""
         return fp * o
 
-    def ln(self, o, fp):
+    @process.register(Ln)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a ln."""
         (f,) = o.ufl_operands
         if isinstance(f, Zero):
             raise ZeroDivisionError()
         return fp / f
 
-    def cos(self, o, fp):
+    @process.register(Cos)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a cos."""
         (f,) = o.ufl_operands
         return fp * -sin(f)
 
-    def sin(self, o, fp):
+    @process.register(Sin)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a sin."""
         (f,) = o.ufl_operands
         return fp * cos(f)
 
-    def tan(self, o, fp):
+    @process.register(Tan)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a tan."""
         (f,) = o.ufl_operands
         return 2.0 * fp / (cos(2.0 * f) + 1.0)
 
-    def cosh(self, o, fp):
+    @process.register(Cosh)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a cosh."""
         (f,) = o.ufl_operands
         return fp * sinh(f)
 
-    def sinh(self, o, fp):
+    @process.register(Sinh)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a sinh."""
         (f,) = o.ufl_operands
         return fp * cosh(f)
 
-    def tanh(self, o, fp):
+    @process.register(Tanh)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a tanh."""
         (f,) = o.ufl_operands
 
@@ -441,34 +547,46 @@ class GenericDerivativeRuleset(MultiFunction):
 
         return fp * sech(f) ** 2
 
-    def acos(self, o, fp):
+    @process.register(Acos)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate an acos."""
         (f,) = o.ufl_operands
         return -fp / sqrt(1.0 - f**2)
 
-    def asin(self, o, fp):
+    @process.register(Asin)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate an asin."""
         (f,) = o.ufl_operands
         return fp / sqrt(1.0 - f**2)
 
-    def atan(self, o, fp):
+    @process.register(Atan)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate an atan."""
         (f,) = o.ufl_operands
         return fp / (1.0 + f**2)
 
-    def atan2(self, o, fp, gp):
+    @process.register(Atan2)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr, gp: Expr) -> Expr:
         """Differentiate an atan2."""
         f, g = o.ufl_operands
         return (g * fp - f * gp) / (f**2 + g**2)
 
-    def erf(self, o, fp):
+    @process.register(Erf)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate an erf."""
         (f,) = o.ufl_operands
         return fp * (2.0 / sqrt(pi) * exp(-(f**2)))
 
     # --- Bessel functions
 
-    def bessel_j(self, o, nup, fp):
+    @process.register(BesselJ)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, nup: Expr, fp: Expr) -> Expr:
         """Differentiate a bessel_j."""
         nu, f = o.ufl_operands
         if not (nup is None or isinstance(nup, Zero)):
@@ -482,7 +600,9 @@ class GenericDerivativeRuleset(MultiFunction):
             op = 0.5 * (bessel_J(nu - 1, f) - bessel_J(nu + 1, f))
         return op * fp
 
-    def bessel_y(self, o, nup, fp):
+    @process.register(BesselY)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, nup: Expr, fp: Expr) -> Expr:
         """Differentiate a bessel_y."""
         nu, f = o.ufl_operands
         if not (nup is None or isinstance(nup, Zero)):
@@ -496,7 +616,9 @@ class GenericDerivativeRuleset(MultiFunction):
             op = 0.5 * (bessel_Y(nu - 1, f) - bessel_Y(nu + 1, f))
         return op * fp
 
-    def bessel_i(self, o, nup, fp):
+    @process.register(BesselI)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, nup: Expr, fp: Expr) -> Expr:
         """Differentiate a bessel_i."""
         nu, f = o.ufl_operands
         if not (nup is None or isinstance(nup, Zero)):
@@ -510,7 +632,9 @@ class GenericDerivativeRuleset(MultiFunction):
             op = 0.5 * (bessel_I(nu - 1, f) + bessel_I(nu + 1, f))
         return op * fp
 
-    def bessel_k(self, o, nup, fp):
+    @process.register(BesselK)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, nup: Expr, fp: Expr) -> Expr:
         """Differentiate a bessel_k."""
         nu, f = o.ufl_operands
         if not (nup is None or isinstance(nup, Zero)):
@@ -526,7 +650,9 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Restrictions
 
-    def restricted(self, o, fp):
+    @process.register(Restricted)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a restricted."""
         # Restriction and differentiation commutes
         if isinstance(fp, ConstantValue):
@@ -536,17 +662,21 @@ class GenericDerivativeRuleset(MultiFunction):
 
     # --- Conditionals
 
-    def binary_condition(self, o, dl, dr):
+    @process.register(BinaryCondition)
+    def _(self, o: Expr) -> Expr:
         """Differentiate a binary_condition."""
         # Should not be used anywhere...
         return None
 
-    def not_condition(self, o, c):
+    @process.register(NotCondition)
+    def _(self, o: Expr) -> Expr:
         """Differentiate a not_condition."""
         # Should not be used anywhere...
         return None
 
-    def conditional(self, o, unused_dc, dt, df):
+    @process.register(Conditional)
+    @DAGTraverser.postorder_only_children([1, 2])
+    def _(self, o: Expr, dt: Expr, df: Expr) -> Expr:
         """Differentiate a conditional."""
         if isinstance(dt, Zero) and isinstance(df, Zero):
             # Assuming dt and df have the same indices here, which
@@ -559,7 +689,9 @@ class GenericDerivativeRuleset(MultiFunction):
             c = o.ufl_operands[0]
             return conditional(c, dt, df)
 
-    def max_value(self, o, df, dg):
+    @process.register(MaxValue)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr, dg: Expr) -> Expr:
         """Differentiate a max_value."""
         # d/dx max(f, g) =
         # f > g: df/dx
@@ -570,7 +702,9 @@ class GenericDerivativeRuleset(MultiFunction):
         dc = conditional(f > g, 1, 0)
         return dc * df + (1.0 - dc) * dg
 
-    def min_value(self, o, df, dg):
+    @process.register(MinValue)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, df: Expr, dg: Expr) -> Expr:
         """Differentiate a min_value."""
         # d/dx min(f, g) =
         #  f < g: df/dx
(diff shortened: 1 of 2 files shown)


ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,120 @@
+"""Base class for dag traversers."""
+
+from abc import ABC, abstractmethod
+from functools import singledispatchmethod, wraps
+
+from ufl.classes import Expr
+
+
+class DAGTraverser(ABC):
+    """Base class for dag traversers."""
```
connorjward, line 10: You should document the `__init__` parameters here

ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __init__(self, compress=True, vcache=None, rcache=None) -> None:
```
connorjward, line 12: I think `rcache` and `vcache` are very obscure names. Could they be made clearer?

ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,120 @@
...
+            # to be able to keep representation of result compact
+            if self._compress:
+                result_ = self._rcache.get(result)
+                if result_ is None:
+                    # Cache miss: store in rcache
+                    self._rcache[result] = result
+                else:
+                    # Cache hit: Use previously computed object result_,
+                    # allowing result to be garbage collected as soon as possible
+                    result = result_
```
connorjward, line 47: Might this be cleaner?
```suggestion
                try:
                    result = self._rcache[result]
                except KeyError:
                    self._rcache[result] = result
```

Also this seems kind of like a `set` to me, not a `dict`.

ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        new_ufl_operands = [self(operand) for operand in o.ufl_operands]
+        if all(nc == c for nc, c in zip(new_ufl_operands, o.ufl_operands)):
+            return o
+        else:
+            return o._ufl_expr_reconstruct_(*new_ufl_operands)
+
+    @staticmethod
+    def postorder(method):
+        """Suppress processed operands in arguments.
```
connorjward, line 85: Could this be clearer?

ufl/algorithms/apply_derivatives.py
```diff
@@ -165,25 +188,65 @@ def independent_operator(self, o):
...
+        """Process ``o``.
+
+        Args:
+            o: `Expr` to be processed.
+
+        Returns:
+            Processed object.
+
+        """
+        raise AssertionError(f"UFL expression expected: got {o}")
```
connorjward, line 204: Is this now boilerplate that needs reproducing for each subclass? Might it be better on the ABC?

ufl/algorithms/apply_derivatives.py
```diff
@@ -257,11 +323,14 @@ def indexed(self, o, Ap, ii):
...
-    def list_tensor(self, o, *dops):
+    @process.register(ListTensor)
+    def _(self, o: Expr) -> Expr:
         """Differentiate a list_tensor."""
-        return ListTensor(*dops)
+        return ListTensor(*(self(op) for op in o.ufl_operands))
 
-    def component_tensor(self, o, Ap, ii):
+    @process.register(ComponentTensor)
+    @DAGTraverser.postorder
```
connorjward, line 332: This is pretty sweet

ufl/algorithms/apply_derivatives.py
```diff
@@ -523,8 +639,11 @@ def bessel_k(self, o, nup, fp):
 
     # --- Restrictions
 
-    def restricted(self, o, fp):
+    @process.register(Restricted)
+    @DAGTraverser.postorder
+    def _(self, o: Expr, fp: Expr) -> Expr:
         """Differentiate a restricted."""
+        (f,) = o.ufl_operands
```
connorjward, line 646: is this line needed?

ufl/algorithms/apply_derivatives.py
```diff
@@ -1435,73 +1716,102 @@ def external_operator(self, N, *dfs):
...
-        # caches for reuse in the dispatched transformers
-        self.vcaches = defaultdict(dict)
-        self.rcaches = defaultdict(dict)
-
+        super().__init__(compress=compress, vcache=vcache, rcache=rcache)
         # Record the operations delayed to the derivative expansion phase:
         # Example: dN(u)/du where `N` is a BaseFormOperator and `u` a Coefficient
         self.pending_operations = ()
+        # Create DAGTraverser caches.
+        self._dag_traverser_dict = {}
```
connorjward, line 1729:
```suggestion
        self._dag_traverser_cache = {}
```

connorjward: and the comment is then unnecessary

ufl/algorithms/apply_derivatives.py
```diff
@@ -1805,38 +2147,61 @@ def jacobian(self, o):
...
     """Dispatcher."""
 
-    def __init__(self):
+    def __init__(self, compress=True, vcache=None, rcache=None) -> None:
         """Initialise."""
-        MultiFunction.__init__(self)
-        self.vcache = defaultdict(dict)
-        self.rcache = defaultdict(dict)
+        super().__init__(compress=compress, vcache=vcache, rcache=rcache)
+        self._dag_traverser_dict = {}
```
connorjward, line 2156:
```suggestion
        self._dag_traverser_cache = {}
```

connorjward (changes requested):
I think this is pretty great. I have written a lot of tree traversals and the `DAGTraverser.postorder` (and friends) solves an issue I've been thinking about.

It's almost a shame that this only lives in UFL.

ufl/corealg/dag_traverser.py
ksagiyam: `vcache -> process_cache` and `rcache -> result_cache`?

ufl/corealg/dag_traverser.py
ksagiyam: I found it hard to keep it short. Please let me know if you have a suggestion. We at least have an expanded explanation right below.

ufl/algorithms/apply_derivatives.py
ksagiyam: Do you mean putting this base case in the base class? I think the problem is that `singledispatchmethod` object needs to be made in each child class. What I had to do here is probably relevant https://github.com/firedrakeproject/ufl/blob/c16f038378e17ad8bd730bf9a020028ad72a00c1/ufl/algorithms/apply_derivatives.py#L716.

ufl/algorithms/apply_derivatives.py
ksagiyam: Removed.

ufl/corealg/dag_traverser.py
connorjward: Or `vcache -> visited_cache` perhaps.

Does `rcache` need to be a dict or can it be a set? In which case it could be `seen_exprs` or `object_cache` or something.

ksagiyam: Changed `vcache -> visited_cache` and `rcache -> result_cache`.

I think `rcache` wants to be a `dict`. If we have `new_o in rcache`, we know that there is some `o` in `rcache` for which `hash(o) == hash(new_o)` and `o == new_o`, but if `rcache` was just a set, I don't see a straightforward way of getting `o` out of `rcache`.

ufl/corealg/dag_traverser.py
ksagiyam: Changed to "Postorder decorator".

ufl/corealg/dag_traverser.py
connorjward: Ah fair enough. It just feels odd to write `cache[result] = result`. But I agree with you now.

And `result_cache` is good.

ksagiyam:
@jorgensd @mscroggs Could you have a look at this PR?

ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,123 @@
+"""Base class for dag traversers."""
+
+from abc import ABC, abstractmethod
+from functools import singledispatchmethod, wraps
+
+from ufl.classes import Expr
+
+
+class DAGTraverser(ABC):
+    """Base class for dag traversers.
```
jorgensd, line 10:
```suggestion
    """Base class for DAG traversers.
```

ufl/algorithms/apply_derivatives.py
```diff
@@ -165,25 +190,65 @@ def independent_operator(self, o):
...
+    @process.register(Expr)
+    def _(self, o: Expr) -> Expr:
+        """Raise error."""
+        raise ValueError(
+            f"Missing differentiation handler for type {o._ufl_class_.__name__}. "
+            "Have you added a new type?"
+        )
+
+    @process.register(Derivative)
+    def _(self, o: Expr) -> Expr:
```
jorgensd, line 217: Wouldn't it be better to use 
```python
@process.register
    def _(self, o: Derivative) -> Expr:
```
here and in subsequent registrations, as https://docs.python.org/3/library/functools.html#functools.singledispatch states:
> To add overloaded implementations to the function, use the register() attribute of the generic function, which can be used as a decorator. For functions annotated with types, the decorator will infer the type of the first argument automatically:

I guess we would have to wait for lower version Python to be 3.11 to do this for unions (https://github.com/python/cpython/pull/30017). However, personally, I think this would be a cleaner solution. What do you think?

ufl/corealg/dag_traverser.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Args:
+        compress: If True, ``result_cache`` will be used.
+        visited_cache: cache of intermediate results; expr -> r = self.process(expr, ...).
+        result_cache: cache of result objects for memory reuse, r -> r.
+
+    """
+
+    def __init__(self, compress=True, visited_cache=None, result_cache=None) -> None:
```
jorgensd, line 19:
```suggestion
    def __init__(self, compress: bool=True, visited_cache=None, result_cache=None) -> None:
```
Currently I do not think there is any test in `ufl` that has a `*args` as a non-empty argument. i.e. is always equal to `args=()`. Could you add a test that has `args` being non-empty?

ufl/algorithms/apply_derivatives.py
```diff
@@ -1802,38 +2162,61 @@ def jacobian(self, o):
...
-    def coefficient_derivative(self, o):
+    @process.register(CoefficientDerivative)
+    def _(self, o: Expr) -> Expr:
         """Apply to a coefficient_derivative."""
         return o
 
-    def coordinate_derivative(self, o, f, w, v, cd):
+    @process.register(CoordinateDerivative)
+    @DAGTraverser.postorder_only_children([0])
+    def _(self, o: Expr, f: Expr) -> Expr:
```
jorgensd, line 2219: I guess a test on coordinate-derivative should yield an `*args` that is non-empty.

ufl/corealg/dag_traverser.py
jorgensd: Should the type-hint for `visited_cache` be `dict[tuple[Expr, ...], Expr]|None`, and similarly for `result_cache`?

ufl/algorithms/apply_derivatives.py
```diff
@@ -582,14 +710,35 @@ def min_value(self, o, df, dg):
 class GradRuleset(GenericDerivativeRuleset):
     """Take the grad derivative."""
 
-    def __init__(self, geometric_dimension):
+    def __init__(self, geometric_dimension, compress=True, visited_cache=None, result_cache=None):
```
jorgensd, line 713: I would add type-hints here as well.

ufl/algorithms/apply_derivatives.py
```diff
@@ -1026,29 +1239,35 @@ class GateauxDerivativeRuleset(GenericDerivativeRuleset):
     D_w[v](e) = d/dtau e(w+tau v)|tau=0.
     """
 
-    def __init__(self, coefficients, arguments, coefficient_derivatives):
+    def __init__(
```
jorgensd, line 1242: Type hints :)

ufl/algorithms/apply_derivatives.py
```diff
@@ -1057,25 +1276,57 @@ def __init__(self, coefficients, arguments, coefficient_derivatives):
...
 
-    def coefficient(self, o):
+    def _process_argument(self, o: Expr) -> Expr:
+        return self.independent_terminal(o)
+
+    @process.register(Coefficient)
+    def _(self, o: Expr) -> Expr:
+        return self._process_coefficient(o)
+
+    def _process_coefficient(self, o: Expr) -> Expr:
```
jorgensd, line 1329: Is the doc-string here strictly correct? (I know it existed prior to the PR).
As `_process_coefficient` supports taking a `BaseFormOperator` or a `Cofunction` it is not really a `Coefficient`.

I guess this goes back to the `GateauxDerivativeRuleset` where any `ExprList` can be taken as input and therefore might be out of scop of this PR.

jorgensd:
In general very positive to this change. Will run the tests later today to check for potential regressions.

jorgensd:
Locally, i did not see a big regression with this PR, however, DOLFINx doesn't have the most complicated forms in the tests.
Maybe @schnellerhase , @jhale or @michalhabera has some more complicated forms in DOLFINy that we could check against.

schnellerhase:
Thanks for the ping. Quick testing on https://github.com/fenics-dolfiny/dolfiny/blob/main/demo/spectral/solid_elasticity_spectral.py - modified test case attached we see the following regression:

```
main 
F: 0.22513866424560547s
J: 0.5185086727142334s

https://github.com/firedrakeproject/ufl.git@449b44e67efc0bda6f5d4cd99c92dbd22dafd146
F: 0.7992544174194336s
J: 4.553445982933044s
```


Modified example

```python
#!/usr/bin/env python3

import time

from mpi4py import MPI
from petsc4py import PETSc

import basix
import dolfinx
import ufl
from dolfinx import default_scalar_type as scalar

import mesh_tube3d_gmshapi as mg
import numpy as np
import plot_tube3d_pyvista as pl

import dolfiny

# Basic settings
name = "solid_elasticity_spectral"
comm = MPI.COMM_WORLD

# Geometry and mesh parameters
r, t, h = 0.04, 0.01, 0.1  # [m]
nr, nt, nh = 16, 5, 8

# Create the regular mesh of a tube with given dimensions
gmsh_model, tdim = mg.mesh_tube3d_gmshapi(name, r, t, h, nr, nt, nh, do_quads=True, order=2)

# Get mesh and meshtags
mesh, mts = dolfiny.mesh.gmsh_to_dolfin(gmsh_model, tdim)

# Get merged MeshTags for each codimension
subdomains, subdomains_keys = dolfiny.mesh.merge_meshtags(mesh, mts, tdim - 0)
interfaces, interfaces_keys = dolfiny.mesh.merge_meshtags(mesh, mts, tdim - 1)

# Define shorthands for labelled tags
surface_lower = interfaces_keys["surface_lower"]
surface_upper = interfaces_keys["surface_upper"]

# Define integration measures
dx = ufl.Measure("dx", domain=mesh, subdomain_data=subdomains, metadata={"quadrature_degree": 4})
ds = ufl.Measure("ds", domain=mesh, subdomain_data=interfaces, metadata={"quadrature_degree": 4})

# Define elements
Ue = basix.ufl.element("P", mesh.basix_cell(), 2, shape=(mesh.geometry.dim,))

# Define function spaces
Uf = dolfinx.fem.functionspace(mesh, Ue)

# Define functions
u = dolfinx.fem.Function(Uf, name="u")

u_ = dolfinx.fem.Function(Uf, name="u_")  # boundary conditions

δu = ufl.TestFunction(Uf)

# Define state as (ordered) list of functions
m, δm = [u], [δu]

# output / visualisation
vorder = mesh.geometry.cmap.degree
uo = dolfinx.fem.Function(dolfinx.fem.functionspace(mesh, ("P", vorder, (3,))), name="u")
so = dolfinx.fem.Function(dolfinx.fem.functionspace(mesh, ("P", vorder)), name="s")  # for output

# Kinematics
F = ufl.Identity(3) + ufl.grad(u)

# Strain measure: from Cauchy strain tensor to squares of principal stretches
c, _ = dolfiny.invariants.eigenstate(F.T * F)  # spectral decomposition of C
c = ufl.as_vector(c)  # squares of principal stretches
c = ufl.variable(c)
# Variation of strain measure (squares of principal stretches)
δc = dolfiny.expression.derivative(c, m, δm)

# Elasticity parameters
E = dolfinx.fem.Constant(mesh, scalar(1.0))  # [MPa]
nu = dolfinx.fem.Constant(mesh, scalar(0.4))  # [-]
mu = E / (2 * (1 + nu))
la = E * nu / ((1 + nu) * (1 - 2 * nu))

# Define boundary stress vector (torque at upper face)
x0 = ufl.SpatialCoordinate(mesh)
n0 = ufl.FacetNormal(mesh)
λ = dolfinx.fem.Constant(mesh, scalar(0.0))
t = ufl.cross(x0 - ufl.as_vector([0.0, 0.0, h]), n0) * 4 * λ  # [N/m^2]


def strain_energy(i1, i2, i3):
    """Strain energy function
    i1, i2, i3: principal invariants of the Cauchy-Green tensor
    """
    # Determinant of configuration gradient F
    J = ufl.sqrt(i3)
    #
    # Classical St. Venant-Kirchhoff
    # Ψ = la / 8 * (i1 - 3) ** 2 + mu / 4 * ((i1 - 3) ** 2 + 4 * (i1 - 3) - 2 * (i2 - 3))
    # Modified St. Venant-Kirchhoff
    # Ψ = la / 2 * (ufl.ln(J)) ** 2 + mu / 4 * ((i1 - 3) ** 2 + 4 * (i1 - 3) - 2 * (i2 - 3))
    # Compressible neo-Hooke
    Ψ = mu / 2 * (i1 - 3 - 2 * ufl.ln(J)) + la / 2 * (J - 1) ** 2
    # Compressible Mooney-Rivlin (beta = 0)
    # Ψ = mu / 4 * (i1 - 3) + mu / 4 * (i2 - 3) - mu * ufl.ln(J) + la / 2 * (J - 1) ** 2
    #
    return Ψ


# Invariants (based on spectral decomposition of C)
i1, i2, i3 = c[0] + c[1] + c[2], c[0] * c[1] + c[1] * c[2] + c[0] * c[2], c[0] * c[1] * c[2]
# Material model (isotropic)
Ψ = strain_energy(i1, i2, i3)
# Stress measure
s = 2 * ufl.diff(Ψ, c)

# von Mises stress (output)
svm = ufl.sqrt(3 / 2 * ufl.inner(ufl.dev(ufl.diag(s)), ufl.dev(ufl.diag(s))))

# Weak form: for isotropic material, eigenprojectors of C and S are identical
form = -0.5 * ufl.inner(δc, s) * dx + ufl.inner(δu, t) * ds(surface_upper)

# Overall form (as list of forms)
forms = dolfiny.function.extract_blocks(form, δm)

form = forms[0]
n = 10
start = time.time()
for _ in range(n):
    ufl.algorithms.compute_form_data(
        form,
        do_apply_function_pullbacks=True,
        do_apply_default_restrictions=True,
        do_apply_geometry_lowering=True,
        do_apply_restrictions=True,
        complex_mode=False,
    )

print(f"F: {(time.time() - start)/n}s")


J_form = ufl.derivative(forms[0], m[0], ufl.TrialFunction(m[0].function_space))

start = time.time()
for _ in range(n):
    ufl.algorithms.compute_form_data(
        J_form,
        do_apply_function_pullbacks=True,
        do_apply_default_restrictions=True,
        do_apply_geometry_lowering=True,
        do_apply_restrictions=True,
        complex_mode=False,
    )

print(f"J: {(time.time() - start)/n}s")
```

jorgensd:
@schnellerhase your example is quite complex.
I've here narrowed it down to a "pure" UFL example (taken bits and pieces from dolfiny)
```bash
import ufl


def invariants_principal(A):
    """Principal invariants of (real-valued) tensor A.
    https://doi.org/10.1007/978-3-7091-0174-2_3
    """
    i1 = ufl.tr(A)
    i2 = (ufl.tr(A) ** 2 - ufl.tr(A * A)) / 2
    i3 = ufl.det(A)
    return i1, i2, i3




def eigenstate3(A):
    """Eigenvalues and eigenprojectors of the 3x3 (real-valued) tensor A.
    Provides the spectral decomposition A = sum_{a=0}^{2} λ_a * E_a
    with (ordered) eigenvalues λ_a and their associated eigenprojectors E_a = n_a^R x n_a^L.

    Note: Tensor A must not have complex eigenvalues!
    """
    if ufl.shape(A) != (3, 3):
        raise RuntimeError(f"Tensor A of shape {ufl.shape(A)} != (3, 3) is not supported!")
    #
    eps = 3.0e-16  # slightly above 2**-(53 - 1), see https://en.wikipedia.org/wiki/IEEE_754
    #
    A = ufl.variable(A)
    #
    # --- determine eigenvalues λ0, λ1, λ2

    # Invariants
    I1, _, _ = invariants_principal(A)

    # Discriminant as sum-of-products
    Δx = [
        A[0, 1] * A[1, 2] * A[2, 0] - A[0, 2] * A[1, 0] * A[2, 1],
        A[0, 1] ** 2 * A[1, 2]
        - A[0, 1] * A[0, 2] * A[1, 1]
        + A[0, 1] * A[0, 2] * A[2, 2]
        - A[0, 2] ** 2 * A[2, 1],
        A[0, 0] * A[0, 1] * A[2, 1]
        - A[0, 1] ** 2 * A[2, 0]
        - A[0, 1] * A[2, 1] * A[2, 2]
        + A[0, 2] * A[2, 1] ** 2,
        A[0, 0] * A[0, 2] * A[1, 2]
        + A[0, 1] * A[1, 2] ** 2
        - A[0, 2] ** 2 * A[1, 0]
        - A[0, 2] * A[1, 1] * A[1, 2],
        A[0, 0] * A[0, 1] * A[1, 2]
        - A[0, 1] * A[0, 2] * A[1, 0]
        - A[0, 1] * A[1, 2] * A[2, 2]
        + A[0, 2] * A[1, 2] * A[2, 1],
        A[0, 0] * A[0, 2] * A[2, 1]
        - A[0, 1] * A[0, 2] * A[2, 0]
        + A[0, 1] * A[1, 2] * A[2, 1]
        - A[0, 2] * A[1, 1] * A[2, 1],
        A[0, 1] * A[1, 0] * A[1, 2]
        - A[0, 2] * A[1, 0] * A[1, 1]
        + A[0, 2] * A[1, 0] * A[2, 2]
        - A[0, 2] * A[1, 2] * A[2, 0],
        A[0, 0] ** 2 * A[1, 2]
        - A[0, 0] * A[0, 2] * A[1, 0]
        - A[0, 0] * A[1, 1] * A[1, 2]
        - A[0, 0] * A[1, 2] * A[2, 2]
        + A[0, 1] * A[1, 0] * A[1, 2]
        + A[0, 2] * A[1, 0] * A[2, 2]
        + A[1, 1] * A[1, 2] * A[2, 2]
        - A[1, 2] ** 2 * A[2, 1],
        A[0, 0] ** 2 * A[1, 2]
        - A[0, 0] * A[0, 2] * A[1, 0]
        - A[0, 0] * A[1, 1] * A[1, 2]
        - A[0, 0] * A[1, 2] * A[2, 2]
        + A[0, 2] * A[1, 0] * A[1, 1]
        + A[0, 2] * A[1, 2] * A[2, 0]
        + A[1, 1] * A[1, 2] * A[2, 2]
        - A[1, 2] ** 2 * A[2, 1],
        A[0, 0] * A[0, 1] * A[1, 1]
        - A[0, 0] * A[0, 1] * A[2, 2]
        - A[0, 1] ** 2 * A[1, 0]
        + A[0, 1] * A[0, 2] * A[2, 0]
        - A[0, 1] * A[1, 1] * A[2, 2]
        + A[0, 1] * A[2, 2] ** 2
        + A[0, 2] * A[1, 1] * A[2, 1]
        - A[0, 2] * A[2, 1] * A[2, 2],
        A[0, 0] * A[0, 1] * A[1, 1]
        - A[0, 0] * A[0, 1] * A[2, 2]
        + A[0, 0] * A[0, 2] * A[2, 1]
        - A[0, 1] ** 2 * A[1, 0]
        - A[0, 1] * A[1, 1] * A[2, 2]
        + A[0, 1] * A[1, 2] * A[2, 1]
        + A[0, 1] * A[2, 2] ** 2
        - A[0, 2] * A[2, 1] * A[2, 2],
        A[0, 0] * A[0, 1] * A[1, 2]
        - A[0, 0] * A[0, 2] * A[1, 1]
        + A[0, 0] * A[0, 2] * A[2, 2]
        - A[0, 1] * A[1, 1] * A[1, 2]
        - A[0, 2] ** 2 * A[2, 0]
        + A[0, 2] * A[1, 1] ** 2
        - A[0, 2] * A[1, 1] * A[2, 2]
        + A[0, 2] * A[1, 2] * A[2, 1],
        A[0, 0] * A[0, 2] * A[1, 1]
        - A[0, 0] * A[0, 2] * A[2, 2]
        - A[0, 1] * A[0, 2] * A[1, 0]
        + A[0, 1] * A[1, 1] * A[1, 2]
        - A[0, 1] * A[1, 2] * A[2, 2]
        + A[0, 2] ** 2 * A[2, 0]
        - A[0, 2] * A[1, 1] ** 2
        + A[0, 2] * A[1, 1] * A[2, 2],
        A[0, 0] ** 2 * A[1, 1]
        - A[0, 0] ** 2 * A[2, 2]
        - A[0, 0] * A[0, 1] * A[1, 0]
        + A[0, 0] * A[0, 2] * A[2, 0]
        - A[0, 0] * A[1, 1] ** 2
        + A[0, 0] * A[2, 2] ** 2
        + A[0, 1] * A[1, 0] * A[1, 1]
        - A[0, 2] * A[2, 0] * A[2, 2]
        + A[1, 1] ** 2 * A[2, 2]
        - A[1, 1] * A[1, 2] * A[2, 1]
        - A[1, 1] * A[2, 2] ** 2
        + A[1, 2] * A[2, 1] * A[2, 2],
    ]
    Δy = [
        A[0, 2] * A[1, 0] * A[2, 1] - A[0, 1] * A[1, 2] * A[2, 0],
        A[1, 0] ** 2 * A[2, 1]
        - A[1, 0] * A[1, 1] * A[2, 0]
        + A[1, 0] * A[2, 0] * A[2, 2]
        - A[1, 2] * A[2, 0] ** 2,
        A[0, 0] * A[1, 0] * A[1, 2]
        - A[0, 2] * A[1, 0] ** 2
        - A[1, 0] * A[1, 2] * A[2, 2]
        + A[1, 2] ** 2 * A[2, 0],
        A[0, 0] * A[2, 0] * A[2, 1]
        - A[0, 1] * A[2, 0] ** 2
        + A[1, 0] * A[2, 1] ** 2
        - A[1, 1] * A[2, 0] * A[2, 1],
        A[0, 0] * A[1, 0] * A[2, 1]
        - A[0, 1] * A[1, 0] * A[2, 0]
        - A[1, 0] * A[2, 1] * A[2, 2]
        + A[1, 2] * A[2, 0] * A[2, 1],
        A[0, 0] * A[1, 2] * A[2, 0]
        - A[0, 2] * A[1, 0] * A[2, 0]
        + A[1, 0] * A[1, 2] * A[2, 1]
        - A[1, 1] * A[1, 2] * A[2, 0],
        A[0, 1] * A[1, 0] * A[2, 1]
        - A[0, 1] * A[1, 1] * A[2, 0]
        + A[0, 1] * A[2, 0] * A[2, 2]
        - A[0, 2] * A[2, 0] * A[2, 1],
        A[0, 0] ** 2 * A[2, 1]
        - A[0, 0] * A[0, 1] * A[2, 0]
        - A[0, 0] * A[1, 1] * A[2, 1]
        - A[0, 0] * A[2, 1] * A[2, 2]
        + A[0, 1] * A[1, 0] * A[2, 1]
        + A[0, 1] * A[2, 0] * A[2, 2]
        + A[1, 1] * A[2, 1] * A[2, 2]
        - A[1, 2] * A[2, 1] ** 2,
        A[0, 0] ** 2 * A[2, 1]
        - A[0, 0] * A[0, 1] * A[2, 0]
        - A[0, 0] * A[1, 1] * A[2, 1]
        - A[0, 0] * A[2, 1] * A[2, 2]
        + A[0, 1] * A[1, 1] * A[2, 0]
        + A[0, 2] * A[2, 0] * A[2, 1]
        + A[1, 1] * A[2, 1] * A[2, 2]
        - A[1, 2] * A[2, 1] ** 2,
        A[0, 0] * A[1, 0] * A[1, 1]
        - A[0, 0] * A[1, 0] * A[2, 2]
        - A[0, 1] * A[1, 0] ** 2
        + A[0, 2] * A[1, 0] * A[2, 0]
        - A[1, 0] * A[1, 1] * A[2, 2]
        + A[1, 0] * A[2, 2] ** 2
        + A[1, 1] * A[1, 2] * A[2, 0]
        - A[1, 2] * A[2, 0] * A[2, 2],
        A[0, 0] * A[1, 0] * A[1, 1]
        - A[0, 0] * A[1, 0] * A[2, 2]
        + A[0, 0] * A[1, 2] * A[2, 0]
        - A[0, 1] * A[1, 0] ** 2
        - A[1, 0] * A[1, 1] * A[2, 2]
        + A[1, 0] * A[1, 2] * A[2, 1]
        + A[1, 0] * A[2, 2] ** 2
        - A[1, 2] * A[2, 0] * A[2, 2],
        A[0, 0] * A[1, 0] * A[2, 1]
        - A[0, 0] * A[1, 1] * A[2, 0]
        + A[0, 0] * A[2, 0] * A[2, 2]
        - A[0, 2] * A[2, 0] ** 2
        - A[1, 0] * A[1, 1] * A[2, 1]
        + A[1, 1] ** 2 * A[2, 0]
        - A[1, 1] * A[2, 0] * A[2, 2]
        + A[1, 2] * A[2, 0] * A[2, 1],
        A[0, 0] * A[1, 1] * A[2, 0]
        - A[0, 0] * A[2, 0] * A[2, 2]
        - A[0, 1] * A[1, 0] * A[2, 0]
        + A[0, 2] * A[2, 0] ** 2
        + A[1, 0] * A[1, 1] * A[2, 1]
        - A[1, 0] * A[2, 1] * A[2, 2]
        - A[1, 1] ** 2 * A[2, 0]
        + A[1, 1] * A[2, 0] * A[2, 2],
        A[0, 0] ** 2 * A[1, 1]
        - A[0, 0] ** 2 * A[2, 2]
        - A[0, 0] * A[0, 1] * A[1, 0]
        + A[0, 0] * A[0, 2] * A[2, 0]
        - A[0, 0] * A[1, 1] ** 2
        + A[0, 0] * A[2, 2] ** 2
        + A[0, 1] * A[1, 0] * A[1, 1]
        - A[0, 2] * A[2, 0] * A[2, 2]
        + A[1, 1] ** 2 * A[2, 2]
        - A[1, 1] * A[1, 2] * A[2, 1]
        - A[1, 1] * A[2, 2] ** 2
        + A[1, 2] * A[2, 1] * A[2, 2],
    ]
    Δd = [9, 6, 6, 6, 8, 8, 8, 2, 2, 2, 2, 2, 2, 1]
    Δ = sum(Δxk * Δdk * Δyk for Δxk, Δdk, Δyk in zip(Δx, Δd, Δy))  # discriminant as sop

    # Invariant dp as sum-of-products
    Δxp = [A[1, 0], A[2, 0], A[2, 1], -A[0, 0] + A[1, 1], -A[0, 0] + A[2, 2], -A[1, 1] + A[2, 2]]
    Δyp = [A[0, 1], A[0, 2], A[1, 2], -A[0, 0] + A[1, 1], -A[0, 0] + A[2, 2], -A[1, 1] + A[2, 2]]
    Δdp = [6, 6, 6, 1, 1, 1]
    dp = sum(Δxpk * Δdpk * Δypk for Δxpk, Δdpk, Δypk in zip(Δxp, Δdp, Δyp)) / 2  # dp as sop

    # Invariant dq as sum-of-products
    Δxq = [
        A[1, 2],
        A[2, 1],
        A[0, 1] * A[1, 0],
        A[0, 2] * A[2, 0],
        A[1, 2] * A[2, 1],
        A[1, 1] + A[2, 2] - 2 * A[0, 0],
    ]
    Δyq = [
        A[0, 1] * A[2, 0],
        A[0, 2] * A[1, 0],
        A[0, 0] + A[1, 1] - 2 * A[2, 2],
        A[0, 0] + A[2, 2] - 2 * A[1, 1],
        A[1, 1] + A[2, 2] - 2 * A[0, 0],
        (A[0, 0] + A[2, 2] - 2 * A[1, 1]) * (A[0, 0] + A[1, 1] - 2 * A[2, 2]),
    ]
    Δdq = [27, 27, 9, 9, 9, -1]
    dq = sum(Δxqk * Δdqk * Δyqk for Δxqk, Δdqk, Δyqk in zip(Δxq, Δdq, Δyq))

    # Avoid dp = 0 and disc = 0, both are known with absolute error of ~eps**2
    # Required to avoid sqrt(0) derivatives and negative square roots
    dp += eps**2
    Δ += eps**2

    phi3 = ufl.atan2(ufl.sqrt(27) * ufl.sqrt(Δ), dq)

    # sorted eigenvalues: λ0 <= λ1 <= λ2
    λ = [(I1 + 2 * ufl.sqrt(dp) * ufl.cos((phi3 + 2 * ufl.pi * k) / 3)) / 3 for k in range(1, 4)]
    #
    # --- determine eigenprojectors E0, E1, E2
    #
    E = [ufl.diff(λk, A).T for λk in λ]

    return λ, E


def eigenstate2(A):
    """Eigenvalues and eigenprojectors of the 2x2 (real-valued) tensor A.
    Provides the spectral decomposition A = sum_{a=0}^{1} λ_a * E_a
    with (ordered) eigenvalues λ_a and their associated eigenprojectors E_a = n_a^R x n_a^L.

    Note: Tensor A must not have complex eigenvalues!
    """
    if ufl.shape(A) != (2, 2):
        raise RuntimeError(f"Tensor A of shape {ufl.shape(A)} != (2, 2) is not supported!")
    #
    eps = 3.0e-16  # slightly above 2**-(53 - 1), see https://en.wikipedia.org/wiki/IEEE_754
    #
    A = ufl.variable(A)
    #
    # --- determine eigenvalues λ0, λ1
    #
    I1, _, _ = invariants_principal(A)
    #
    Δ = (A[0, 0] - A[1, 1]) ** 2 + 4 * A[0, 1] * A[1, 0]  # = I1**2 - 4 * I2
    # Avoid dp = 0 and disc = 0, both are known with absolute error of ~eps**2
    # Required to avoid sqrt(0) derivatives and negative square roots
    Δ += eps**2
    # sorted eigenvalues: λ0 <= λ1
    λ = (I1 - ufl.sqrt(Δ)) / 2, (I1 + ufl.sqrt(Δ)) / 2
    #
    # --- determine eigenprojectors E0, E1
    #
    E = [ufl.diff(λk, A).T for λk in λ]

    return λ, E


def eigenstate(A):
    """Eigenvalues and eigenprojectors of the (real-valued) tensor A of dimension m = 2, 3.
    Provides the spectral decomposition A = sum_{a=0}^{m} λ_a * E_a
    with (ordered) eigenvalues λ_a and their associated eigenprojectors E_a = n_a^R x n_a^L.

    Note: Tensor A must not have complex eigenvalues!
    """
    if ufl.shape(A) == (3, 3):
        return eigenstate3(A)
    elif ufl.shape(A) == (2, 2):
        return eigenstate2(A)
    else:
        raise RuntimeError(f"Tensor A of shape {ufl.shape(A)} is not supported!")



from ufl import Coefficient, FunctionSpace, Mesh, TestFunction, TrialFunction, dx, hexahedron, Constant

element = FiniteElement("Lagrange", hexahedron, 2, (3,), identity_pullback, H1)
domain = Mesh(FiniteElement("Lagrange", hexahedron, 2, (3,), identity_pullback, H1))


# Define integration measures
dx = ufl.Measure("dx", domain=domain, metadata={"quadrature_degree": 4})
ds = ufl.Measure("ds", domain=domain, metadata={"quadrature_degree": 4})


Uf = FunctionSpace(domain, element)
u = Coefficient(Uf)  # displacement
u_ = Coefficient(Uf)

δu = TestFunction(Uf)


# output / visualisation

# Kinematics
F = ufl.Identity(3) + ufl.grad(u)

# Strain measure: from Cauchy strain tensor to squares of principal stretches
c, _ = eigenstate(F.T * F)  # spectral decomposition of C
c = ufl.as_vector(c)  # squares of principal stretches
c = ufl.variable(c)
# Variation of strain measure (squares of principal stretches)
δc = ufl.derivative(c, u, δu)

# Elasticity parameters
E = Constant(domain)  # [MPa]
nu = Constant(domain)
mu = E / (2 * (1 + nu))
la = E * nu / ((1 + nu) * (1 - 2 * nu))

# Define boundary stress vector (torque at upper face)
x0 = ufl.SpatialCoordinate(domain)
n0 = ufl.FacetNormal(domain)
λ = Constant(domain)
h = Constant(domain)  # [m]
t = ufl.cross(x0 - ufl.as_vector([0.0, 0.0, h]), n0) * 4 * λ  # [N/m^2]


def strain_energy(i1, i2, i3):
    """Strain energy function
    i1, i2, i3: principal invariants of the Cauchy-Green tensor
    """
    # Determinant of configuration gradient F
    J = ufl.sqrt(i3)
    #
    # Classical St. Venant-Kirchhoff
    # Ψ = la / 8 * (i1 - 3) ** 2 + mu / 4 * ((i1 - 3) ** 2 + 4 * (i1 - 3) - 2 * (i2 - 3))
    # Modified St. Venant-Kirchhoff
    # Ψ = la / 2 * (ufl.ln(J)) ** 2 + mu / 4 * ((i1 - 3) ** 2 + 4 * (i1 - 3) - 2 * (i2 - 3))
    # Compressible neo-Hooke
    Ψ = mu / 2 * (i1 - 3 - 2 * ufl.ln(J)) + la / 2 * (J - 1) ** 2
    # Compressible Mooney-Rivlin (beta = 0)
    # Ψ = mu / 4 * (i1 - 3) + mu / 4 * (i2 - 3) - mu * ufl.ln(J) + la / 2 * (J - 1) ** 2
    #
    return Ψ


# Invariants (based on spectral decomposition of C)
i1, i2, i3 = c[0] + c[1] + c[2], c[0] * c[1] + c[1] * c[2] + c[0] * c[2], c[0] * c[1] * c[2]
# Material model (isotropic)
Ψ = strain_energy(i1, i2, i3)
# Stress measure
s = 2 * ufl.diff(Ψ, c)

# von Mises stress (output)
svm = ufl.sqrt(3 / 2 * ufl.inner(ufl.dev(ufl.diag(s)), ufl.dev(ufl.diag(s))))

# Weak form: for isotropic material, eigenprojectors of C and S are identical
form = -0.5 * ufl.inner(δc, s) * dx + ufl.inner(δu, t) * ds(1)


import time
start = time.perf_counter()
ufl.algorithms.compute_form_data(
    form,
    do_apply_function_pullbacks=True,
    do_apply_default_restrictions=True,
    do_apply_geometry_lowering=True,
    do_apply_restrictions=True,
    complex_mode=False,
)
end = time.perf_counter()
print(f"Time taken to compute form data (residual): {(end - start):.2e}")

J = ufl.derivative(form, u, δu)

start = time.perf_counter()
ufl.algorithms.compute_form_data(
    form,
    do_apply_function_pullbacks=True,
    do_apply_default_restrictions=True,
    do_apply_geometry_lowering=True,
    do_apply_restrictions=True,
    complex_mode=False,
)
end = time.perf_counter()
print(f"Time taken to compute form data (jacobian): {(end - start):.2e}")
```
which results in the following timings on main:
```bash
Time taken to compute form data (residual): 1.65e-01
Time taken to compute form data (jacobian): 1.16e-01
```
and the following on this branch:
```bash
Time taken to compute form data (residual): 6.21e-01
Time taken to compute form data (jacobian): 5.29e-01
```

jorgensd:
There was a small typo in the example above (didn't compile form data for jacobian).
This has now been updated, and the timings have been updated, showing the 10-fold increase in runtime for computing form data for the jacobian.

ksagiyam:
Thanks! I had not updated my branch after some performance fix was merged to `main`. After rebasing, the numbers are like the following:

`main`:
Time taken to compute form data (residual): 8.51e-02
Time taken to compute form data (jacobian): 1.62e-01

`this PR`:
Time taken to compute form data (residual): 1.01e-01 (+19%)
Time taken to compute form data (jacobian): 2.01e-01 (+24%)

Let me also update the results for the other tests in the PR description:

`holzapfel_ogden`:

`main`:
0.04760003089904785

`this PR`:
0.05493581295013428 (+15%)

`wence_test`:

`main`:
0.06049168109893799

`this PR`:
0.06576049327850342 (+8.7%)

ufl/corealg/dag_traverser.py
ksagiyam: Added typehint.

That is right. It seems that we only need `args = ()` in `apply_derivatives.py`. We will need non-empty `args` for instance when we want to pass down the state in `apply_restrictions.py`; underneath a `Restricted` node `restricted='+'` or `restricted='-'`, otherwise `restricted=None`, where the method signature would look like:
```
@process.register(...)
def _(self, o, restricted):
    ...
```
We could remove `*args` for this PR.

ufl/algorithms/apply_derivatives.py
ksagiyam: `CoordinateDerivative ` still only needs `args=()`, I think.

ufl/algorithms/apply_derivatives.py
ksagiyam: Done.

ufl/algorithms/apply_derivatives.py
ksagiyam: Done.

ufl/algorithms/apply_derivatives.py
ksagiyam: I partially fixed this (as I added these new methods), but I think we should do more comprehensive fix in a future PR.

ufl/algorithms/apply_derivatives.py
ksagiyam: I gave it a go with this as a separate commit, but I personally prefer:
```
    @process.register(Grad)
    @process.register(CellAvg)
    @process.register(FacetAvg)
    def _(self, o) -> Expr:
        ...
```
as type names are left-aligned and easier to read for me. I also do not quite see how it plays well when decorators are nested as in:
```
    @process.register
    @DAGTraverser.postorder
    @pending_operations_recording
    def external_operator(self, N: ExternalOperator, *dfs) -> Expr:
        ...
```
though `functools.wraps()` seems to be doing the job for us.

Please see the history for changes; I would like to revert this change back for now as Python3.10 test fails as you already pointed out.

ufl/algorithms/apply_derivatives.py
jorgensd: Yeah, I only looked at the signatures of the process method, but for the traversal of the graph, it seems like we never use args.

ufl/algorithms/apply_derivatives.py
dham: maybe call super like you do in the other case.

ufl/algorithms/apply_derivatives.py
dham: I think we should follow the Python convention here, which is the way Jørgen has suggested.

ufl/algorithms/apply_derivatives.py
ksagiyam: Done.

ufl/algorithms/apply_derivatives.py
ksagiyam: Ok, it is good, but we indeed need Python >= 3.11, so we probably should not do this in this PR. I just dropped the relevant commit.

mscroggs: We could make a note of this in https://github.com/FEniCS/ufl/issues/176 to remind us about this in October 2026

ksagiyam: Yes. I will drop a note once this PR is merged.

mscroggs (approved):

PR #367, opened 2025-04-01
remove_compoment_tensors: fix performance regression

pbrubeck:
This PR makes `remove_component_tensors` introduced in https://github.com/FEniCS/ufl/pull/339 an optional step in `compute_form_data`, and disables it by default. Moreover, we also fix a performance regeression (the tree was exploding as unchanged nodes were recreated).

This preprocessing step was creating new multi indices even if they didn't change, which made the dag traversal very expensive. 

Here we introduce reuse of unchanged multi indices and shared caching for the internal rules dispatched by `IndexRemover`. 

We also add a few optimizations to exit early and memoize partial results for equality and comparison of expressions.

### DIFF
--- a/ufl/exprequals.py
+++ b/ufl/exprequals.py
@@ -20,14 +20,15 @@ def expr_equals(self, other):
         return False
 
     # Large objects are costly to compare with themselves
-    if self is other:
+    if self is other or self.ufl_operands is other.ufl_operands:
         return True
 
     # Modelled after pre_traversal to avoid recursion:
     left = [(self, other)]
+    equal_pairs = set()
     while left:
-        s, o = left.pop()
-
+        pair = left.pop()
+        s, o = pair
         if s._ufl_is_terminal_:
             # Compare terminals
             if not s == o:
@@ -36,6 +37,9 @@ def expr_equals(self, other):
             # Delve into subtrees
             so = s.ufl_operands
             oo = o.ufl_operands
+            # Skip subtrees if operands are the same
+            if so is oo:
+                continue
             if len(so) != len(oo):
                 return False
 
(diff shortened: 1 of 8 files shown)


ufl/exprequals.py
```diff
@@ -44,11 +48,16 @@ def expr_equals(self, other):
...
-                if s is o:
+                if s is o or s.ufl_operands is o.ufl_operands:
+                    continue
+                if (s, o) in equal_pairs:
                     continue
                 # Append subtree for further inspection
                 left.append((s, o))
 
+        # Keep track of equal subexpressions
+        equal_pairs.add(pair)
```
dham, line 59: use (id(pair[0]), id(pair[1]))

ufl/exprequals.py
```diff
@@ -44,11 +48,16 @@ def expr_equals(self, other):
                 if s._ufl_typecode_ != o._ufl_typecode_:
                     return False
                 # Skip subtree if objects are the same
-                if s is o:
+                if s is o or s.ufl_operands is o.ufl_operands:
+                    continue
+                if (s, o) in equal_pairs:
```
pbrubeck, line 53:
```suggestion
                if (id(s), id(o)) in equal_pairs:
```

ufl/exprequals.py
pbrubeck:
```suggestion
        equal_pairs.add((id(pair[0]), id(pair[1])))
```

ufl/sorting.py
```diff
@@ -158,6 +164,9 @@ def cmp_expr(a, b):
             if x != y:
                 return -1 if x < y else 1
 
+        # Keep track of equal subexpressions
+        equal_pairs.add(pair)
```
pbrubeck, line 168:
```suggestion
        equal_pairs.add((id(pair[0]), id(pair[1])))
```

ufl/sorting.py
```diff
@@ -139,12 +140,17 @@ def cmp_expr(a, b):
...
+            # Skip subtrees if operands are the same
+            if aops is bops:
+                continue
 
             # Sort by children in natural order
             for r, s in zip(aops, bops):
                 # Skip subtree if objects are the same
                 if r is s:
                     continue
+                if (r, s) in equal_pairs:
```
pbrubeck, line 152:
```suggestion
                if (id(r), id(s)) in equal_pairs:
```

PR #370, opened 2025-04-04
Bug fixes for BaseFormOperator

pbrubeck:


pbrubeck:
Initially `BaseFormOperators` were only meant to be primal operators, am I right @dham? At least `Interpolate.ufl_function_space()` always returned a primal space. They inherit from `Operator` (a primal ufl.Expr) and `BaseForm` (a dual).

BaseFormOperators could be primal or dual objects, depending on their arguments. For instance, `Interpolate(TestFunction(V1), Cofunction(V2.dual()))` lives in `V1.dual()`. On the other hand, `Interpolate(Function(V1), Coargument(V2.dual()))` lives in `V2`.

The problem is that linear combinations of dual objects need a different code path (`FormSum`). How should we handle multiple inheritance?

ufl/core/base_form_operator.py
```diff
@@ -182,4 +183,31 @@ def __hash__(self):
...
+        )
+
+    @property
+    def _parent_type(self):
+        """Is this a primal or dual expression?"""
+        return BaseForm if is_dual(self.ufl_function_space()) else Operator
+
+    def __add__(self, other):
+        """Add."""
+        return self._parent_type.__add__(self, other)
```
pbrubeck, line 201: Is this the right way of dealing with multiple inheritance dynamically?

dham: I think this is an OK thing to do.

ufl/action.py
```diff
@@ -69,6 +70,17 @@ def __new__(cls, *args, **kw):
         if isinstance(right, (Coargument, Argument)):
             return left
 
+        # Simplify Action(Interpolate(Argument, Coargument), BaseForm)
```
dham, line 73: Check argument compatibility before simplifying.

ufl/action.py
```diff
@@ -69,6 +70,17 @@ def __new__(cls, *args, **kw):
...
+        # -> Interpolate(Argument, BaseForm)
+        if (
+            isinstance(left, Interpolate)
+            and len(left.arguments()) == 2
+            and isinstance(right, BaseForm)
+            and len(right.arguments()) == 1
+        ):
+            _, operand = left.argument_slots()
```
dham, line 81: I think this needs to account for transpose interpolate.

ufl/core/base_form_operator.py
```diff
@@ -182,4 +183,31 @@ def __hash__(self):
...
     def __eq__(self, other):
         """Check for equality."""
-        raise NotImplementedError()
+        if self is other:
+            return True
+        return (
+            type(self) is type(other)
+            and all(a == b for a, b in zip(self._argument_slots, other._argument_slots))
+            and self.ufl_function_space() == other.ufl_function_space()
+        )
```
dham, line 192: Do this on Action and Adjoint rather than here.

dham (changes requested):

ufl/core/base_form_operator.py
```diff
@@ -182,4 +184,27 @@ def __hash__(self):
 
     def __eq__(self, other):
         """Check for equality."""
+        if isinstance(other, Number) and other == 0:
+            return self.empty()
         raise NotImplementedError()
```
pbrubeck, line 189: @dham in [adjoint.py](https://github.com/FEniCS/ufl/blob/02cc03a7ce5f9589e1d92639044b3f33830ad7d7/ufl/adjoint.py#L44) we check whether the BaseForm that we are adjoining is zero. In particular, we need this for `BaseFormOperatorDerivative`. Could we implement equality against 0 generically in `BaseFormOperator`?

dham: Quite possibly.

ufl/action.py
pbrubeck: Done

ufl/adjoint.py
```diff
@@ -61,6 +62,15 @@ def __new__(cls, *args, **kw):
...
             return primal_arg
+        elif isinstance(form, Interpolate):
+            from ufl.algorithms.replace import replace
+
+            v, u = form.arguments()
+            mapping = {
+                v: type(v)(v.ufl_function_space(), number=1, part=v.part()),
+                u: type(u)(u.ufl_function_space(), number=0, part=u.part()),
+            }
+            return replace(form, mapping)
```
pbrubeck, line 73: We have been allowing `form = Interpolate(Argument(V1, 0), Coargument(V2, 0))` with both arguments with `number() == 0`. 

As a consequence, `Adjoint(form)` gets the numbering wrong, but this simplification seems to do the right thing, as long as `v, u = form.arguments()` come in the right order.

pbrubeck: I think I solved this issue by eagerly simplifying the `Action` of an `Interpolate` with a `BaseForm`, preventing as `Interpolate` with an `Argument` that eventually gets contracted.

dham (changes requested):

ufl/action.py
```diff
@@ -85,6 +86,31 @@ def __new__(cls, *args, **kw):
...
+
+        # Simplify Action(Interpolate(Expr, Coargument), BaseForm)
+        # -> Interpolate(Expr, BaseForm)
+        if (
+            isinstance(left, Interpolate)
+            and isinstance(right, BaseForm)
+            and len(right.arguments()) == 1
+        ):
+            v, operand = left.argument_slots()
+            if v.number() == len(left.arguments()) - 1:
```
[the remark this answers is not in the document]
pbrubeck:
```suggestion
            if v == left.arguments()[-1]:
```

dham (approved):

PR #371, opened 2025-04-10
GateauxDerivativeRuleset return zero of the correct shape

pbrubeck:
Fixes https://github.com/FEniCS/dolfinx/issues/3698 by reverting changes introduced in https://github.com/FEniCS/ufl/pull/367, namely
- Fast exit logic in exprequals.py, presumably it was wrong because types were not compared.
- Simplification of Indexed Conditional, the issue has to do with derivatives creating Zero of the wrong shape, and this PR is an attempt to fix that.

### DIFF
--- a/ufl/exprequals.py
+++ b/ufl/exprequals.py
@@ -37,6 +37,9 @@ def expr_equals(self, other):
             # Delve into subtrees
             so = s.ufl_operands
             oo = o.ufl_operands
+            # Skip subtree if operands are the same
+            if so is oo:
+                continue
             if len(so) != len(oo):
                 return False
 
--- a/ufl/exprequals.py
+++ b/ufl/exprequals.py
@@ -20,7 +20,7 @@ def expr_equals(self, other):
         return False
 
     # Large objects are costly to compare with themselves
-    if self is other or self.ufl_operands is other.ufl_operands:
+    if (self is other) or (self.ufl_operands is other.ufl_operands):
         return True
 
     # Modelled after pre_traversal to avoid recursion:
@@ -37,9 +37,6 @@ def expr_equals(self, other):
             # Delve into subtrees
             so = s.ufl_operands
             oo = o.ufl_operands
-            # Skip subtrees if operands are the same
-            if so is oo:
-                continue
             if len(so) != len(oo):
                 return False
 
@@ -48,7 +45,7 @@ def expr_equals(self, other):
                 if s._ufl_typecode_ != o._ufl_typecode_:
                     return False
                 # Skip subtree if objects are the same
-                if s is o or s.ufl_operands is o.ufl_operands:
+                if s is o:
                     continue
                 if (id(s), id(o)) in equal_pairs:
                     continue
--- a/ufl/sorting.py
+++ b/ufl/sorting.py
@@ -143,6 +143,9 @@ def cmp_expr(a, b):
             # Delve into subtrees
             aops = a.ufl_operands
             bops = b.ufl_operands
+            # Skip subtree if operands are the same
+            if aops is bops:
+                continue
 
             # Sort by children in natural order
             for r, s in zip(aops, bops):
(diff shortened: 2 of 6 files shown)


schnellerhase:
Can confirm this fixes the reported problem which was a simplified version of https://github.com/fenics-dolfiny/dolfiny/blob/3682843cc8ccf28edd990d763ed0c5487dc35d77/demo/plasticity/solid_plasticity_monolithic.py. 

However we are facing two more failures after the previous PR in https://github.com/fenics-dolfiny/dolfiny/blob/3682843cc8ccf28edd990d763ed0c5487dc35d77/demo/continuation/continuation_planartruss_disp.py and https://github.com/fenics-dolfiny/dolfiny/blob/3682843cc8ccf28edd990d763ed0c5487dc35d77/demo/beam/beam_curved_finitestrain_bstar.py - which persist with these changes. Don't have simplified versions for those yet.

pbrubeck:
> However we are facing two more failures after the previous PR in https://github.com/fenics-dolfiny/dolfiny/blob/3682843cc8ccf28edd990d763ed0c5487dc35d77/demo/continuation/continuation_planartruss_disp.py and https://github.com/fenics-dolfiny/dolfiny/blob/3682843cc8ccf28edd990d763ed0c5487dc35d77/demo/beam/beam_curved_finitestrain_bstar.py - which persist with these changes. Don't have simplified versions for those yet.

I updated the fix, not sure if it could further solve the other two failures. It would be helpful to see the full Traceback.

schnellerhase:
State [3f8bc80](https://github.com/FEniCS/ufl/pull/371/commits/3f8bc80f76679ab56728099d74a2c143ca63f186) works - however [39116ec](https://github.com/FEniCS/ufl/pull/371/commits/39116eca134abd64cf7572500d934281fa390a4c) fails also on the first example.

pbrubeck:
> State [3f8bc80](https://github.com/FEniCS/ufl/pull/371/commits/3f8bc80f76679ab56728099d74a2c143ca63f186) works - however [39116ec](https://github.com/FEniCS/ufl/pull/371/commits/39116eca134abd64cf7572500d934281fa390a4c) fails also on the first example (fixes the MFE however)

How does it fail. Could you share the error message?

schnellerhase:
I've setup a simple CI to debug at https://github.com/schnellerhase/dolfiny/pull/1.

Will add the other examples to it as well. Output is however not too descriptive.

schnellerhase:
Thanks for your efforts @pbrubeck - this fixes our problems.

ufl/sorting.py
```diff
@@ -143,9 +143,6 @@ def cmp_expr(a, b):
             # Delve into subtrees
             aops = a.ufl_operands
             bops = b.ufl_operands
-            # Skip subtrees if operands are the same
-            if aops is bops:
-                continue
```
jorgensd, line 148: Same comment here,  wouldn't this first (removed) check just resolve the comparison quicker?

ufl/exprequals.py
```diff
@@ -37,9 +37,6 @@ def expr_equals(self, other):
             # Delve into subtrees
             so = s.ufl_operands
             oo = o.ufl_operands
-            # Skip subtrees if operands are the same
-            if so is oo:
-                continue
```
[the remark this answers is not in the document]
pbrubeck: You are right, this check all fine, so I'm putting it back in. 

We only add pairs if the types are equal. If we have equal types, we have equality for the pair if the operands are also equal.

ufl/exprequals.py
```diff
@@ -48,7 +48,7 @@ def expr_equals(self, other):
                 if s._ufl_typecode_ != o._ufl_typecode_:
                     return False
                 # Skip subtree if objects are the same
-                if s is o or s.ufl_operands is o.ufl_operands:
```
pbrubeck, line 51: The operand comparison here seems to be the main issue. Removing this one makes the test pass.

Anyways, if a pair in the subtree with equal operands gets added to the stack, we can exit early on it with the check above.

jorgensd: @schnellerhase could you test this ?:)

jorgensd (approved):

Issue #386, opened 2025-06-26
Fix `mypy`'s `attr-defined` errors


schnellerhase:
After introduction of typing support with https://github.com/FEniCS/ufl/pull/356 the `attr-defined` error is the only remaining error which needs to be fixed. 

This will require some work, should also prepare the removal of the internal type system.

PR #387, opened 2025-06-26
Fix `attr-defined` `mypy` errors

schnellerhase:
Fixes mypy `attr-defined` errors, mainly by removing automatic population of the namespaces. 

This also uncovered quite a few indirect `mypy` errors.

There is a more complicated typing problem inside of `apply_derivates` between the usage of `BaseForm` and `Expr`. This appears to require some further fixing on its own. This is currently deactivated by temporarily adding quite a few `ignore`'s to `apply_derivatives.py`. 

Up to this ufl is then mypy compliant. We ignore no more errors on a global scope.

Fixes https://github.com/FEniCS/ufl/issues/386.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # UFL - Unified Form Language
 
+[![UFL CI](https://github.com/FEniCS/ufl/actions/workflows/pythonapp.yml/badge.svg)](https://github.com/FEniCS/ufl/actions/workflows/pythonapp.yml)
+[![Spack build](https://github.com/FEniCS/ufl/actions/workflows/spack.yml/badge.svg)](https://github.com/FEniCS/ufl/actions/workflows/spack.yml)
+[![Coverage Status](https://coveralls.io/repos/github/FEniCS/ufl/badge.svg?branch=main)](https://coveralls.io/github/FEniCS/ufl?branch=main)
+
 The Unified Form Language (UFL) is a domain specific language for
 declaration of finite element discretizations of variational forms. More
 precisely, it defines a flexible interface for choosing finite element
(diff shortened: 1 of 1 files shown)


ufl/algorithms/apply_derivatives.py
```diff
@@ -1342,28 +1342,28 @@ def _(self, o: Expr, fp: Expr) -> Expr:
...
+    def _(self, o: Coefficient) -> Expr:
         return self._process_coefficient(o)
 
     def _process_coefficient(self, o: Union[Expr]) -> Union[Expr]:
         """Differentiate an Expr or a BaseForm."""
         # Define dw/dw := d/ds [w + s v] = v
 
         # Return corresponding argument if we can find o among w
-        do = self._w2v.get(o)
+        do = self._w2v.get(o)  # type: ignore
```
schnellerhase, line 1361:
```suggestion
        do = self._w2v.get(o)
```

ufl/algorithms/apply_derivatives.py
```diff
@@ -1342,28 +1342,28 @@ def _(self, o: Expr, fp: Expr) -> Expr:
...
         if do is not None:
             return do
 
         # Look for o among coefficient derivatives
-        dos = self._cd.get(o)
+        dos = self._cd.get(o)  # type: ignore
```
schnellerhase, line 1366:
```suggestion
        dos = self._cd.get(o)
```

ufl/algorithms/apply_derivatives.py
```diff
@@ -2156,7 +2156,7 @@ def _(self, o: Expr) -> Expr:
     @process.register(SpatialCoordinate)
     def _(self, o: Expr) -> Expr:
         """Differentiate a spatial_coordinate."""
-        do = self._w2v.get(o)
+        do = self._w2v.get(o)  # type: ignore
```
schnellerhase, line 2159:
```suggestion
        do = self._w2v.get(o)
```

mscroggs (approved):

PR #388, opened 2025-06-27
Fix bug in operator derivative reconstruction

indiamai:
Bug description:
 - Firedrake tests failed due to incorrect argument error.
 - Identified root cause as change to how `BaseFormOperatorDerivative` reconstruction is called.

Fix description:
 - Add `self` to the call of reconstruction operator to ensure correct arguments are passed.

Also added a rudimentary test to catch this specific case.


mscroggs (approved):

schnellerhase:
Thanks for the catch. Looks good to me. Should we maybe also raise here, if anything but `None` is passed into the call (which we then drop here)?

PR #390, opened 2025-06-27
Extend `mypy` testing to tests and demos

schnellerhase:
Extends the `mypy` checks to also run across the unit tests and demo files.

Removes no longer supported `_TensorProductElement` demo.


demo/_TensorProductElement.py
```diff
@@ -39,9 +39,9 @@
 
 V = TensorProductElement(V0, V1, V2)
 
-c0 = FiniteElement("CG", triangle, 1)
-c1 = FiniteElement("CG", interval, 1)
-c2 = FiniteElement("CG", tetrahedron, 1)
+c0 = FiniteElement("CG", triangle, 1, (), identity_pullback, L2)
+c1 = FiniteElement("CG", interval, 1, (), identity_pullback, L2)
+c2 = FiniteElement("CG", tetrahedron, 1, (), identity_pullback, L2)
```
schnellerhase, line 44: @mscroggs that correct? And how could that have passed before if those are required arguments for `utils.FiniteElement`?

mscroggs: Demos that start with `_` are skipped in testing so who knows how long this one hasn't worked for https://github.com/FEniCS/ufl/blob/b9dd3250754af58e5e344ffc76aa215da5fba198/test/test_analyse_demos.py#L16

mscroggs (commented):
embedded_superdegree can return None

ufl/finiteelement.py
```diff
@@ -62,7 +62,7 @@ def pullback(self) -> _AbstractPullback:
         """Return the pullback for this element."""
 
     @_abc.abstractproperty
-    def embedded_superdegree(self) -> int | None:
+    def embedded_superdegree(self) -> int:
```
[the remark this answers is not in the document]
schnellerhase: Need to stick with `Optional` for the time being...

PR #400, opened 2025-08-08
Remove constraint on `ReferenceGrad` operand type

schnellerhase:
Removes the type restriction of `ReferenceGrad` operands  to `ReferenceGrad` or `ReferenceValue`.

Minimal reproducible example failing on `main` (this requires an additional `Jacobian` type):

```Python 
from utils import LagrangeElement

import ufl

# Only happens for higher order geometry, changing to degree 1 here yields working example.
domain = ufl.Mesh(LagrangeElement(ufl.interval, 2, (1,)))

J0 = ufl.geometry.Jacobian(domain)

J = ufl.grad(J0)

V = ufl.FunctionSpace(domain, LagrangeElement(ufl.interval, 1))
u = ufl.Coefficient(V)

breaks = ufl.diff(J, u)
breaks = ufl.algorithms.apply_derivatives.apply_derivatives(breaks)
# Breaks down with
# > ValueError: Unexpected argument to reference_grad.
```

Other examples for more types can be created. Second test case hits `SpatialCoordinate`.


schnellerhase:
Possibly resolves https://github.com/firedrakeproject/firedrake/issues/2595 and https://github.com/firedrakeproject/firedrake/issues/3313.

test/test_apply_derivatives.py
```diff
@@ -0,0 +1,130 @@
...
+    ],
+)
+@pytest.mark.parametrize("order", [2, 3])
+@pytest.mark.parametrize("lower_alg", [True, False])
+@pytest.mark.parametrize("lower_geo", [True, False])
+@pytest.mark.parametrize("apply_deriv", [True, False])
+@pytest.mark.skipif(sys.version_info >= (3, 14), reason="Hits recursion error on Python 3.14.0rc1 ")
+def test_diff_grad_grad_jacobian(cell, gdim, order, lower_alg, lower_geo, apply_deriv):
+    tdim = cell.topological_dimension()
+
```
michalhabera, line 96:
```suggestion
```

michalhabera (commented):
There are tests in https://github.com/FEniCS/ufl/blob/main/test/test_automatic_differentiation.py, could you move there? `ExpressionCollection` has geometry objects ready, also some shape checks before/after are in place.

schnellerhase:
Added parametrisation of ad test cases over geometry order and geometric dimension. This should in test `test_grad_coeff` hit our setup. However the test also pass on `main` (thus not hitting the issue at hand). The varied choices of lowering operations are important here to uncover, so the second test case is not merged into the present ones.

dham (approved):

michalhabera (approved):

PR #401, opened 2025-08-11
Removal of custom type system

schnellerhase:


.github/workflows/fenicsx-tests.yml
```diff
@@ -5,7 +5,7 @@ name: FEniCSx integration
 on:
   pull_request:
     branches:
-      - main
+      # - main
```
mscroggs, line 8: Is this change (and the similar change to the TSFC tests) intended

mscroggs (commented):
I'll review this thoroughly at some point but looks like excellent progress towards cleaning up the type system

PR #404, opened 2025-08-13
Update pyproject license to new format

chrisrichardson:


connorjward:
I recommend bumping the setuptools minimum version at the same time as this new format is only for fairly new setuptools. We hit this in Firedrake: https://github.com/firedrakeproject/firedrake/commit/43211b9eb4bb4ee983d0b3da046e3af2754a6efd

PR #412, opened 2025-08-28
`UniqueDomainExtractor` DAG traverser

leo-collins:
Extract unique domain from an expression by traversing the expression tree. If we reach a terminal, we return the domain of that. If we reach an operator, we return the domain of the operands, if they all match.

Unlike the previous `unique_domain_extractor`, the DAG traverser works with expressions containing indexed arguments and coefficients. For example,

```
domain = MeshSequence([mesh1, mesh2])
V = FunctionSpace(domain, mixed_element)
u1, u2 = split(TrialFunction(V))
assert extract_unique_domain(u1) == mesh1
assert extract_unique_domain(u2) == mesh2
```

I plan to extend this in the future so that it works with `BaseForm`s containing indexed arguments and coefficients, though I think this will require implementing an `IndexedBaseForm` class

### DIFF
--- a/test/test_unique_domain_extractor.py
+++ b/test/test_unique_domain_extractor.py
@@ -58,7 +58,7 @@ def test_extract_unique_domain():
     assert extract_unique_domain(expr2) == mesh1
 
     x2, y2 = SpatialCoordinate(mesh2)
-    with pytest.raises(ValueError) as e_info:
+    with pytest.raises(ValueError):
         _ = extract_unique_domain(u1 + u2)
         _ = extract_unique_domain(u1 + u2 + x2 * cos(x2 * u1))
 
--- a/ufl/domain.py
+++ b/ufl/domain.py
@@ -433,6 +433,8 @@ def find_geometric_dimension(expr):
 
 
 class UniqueDomainExtractor(DAGTraverser):
+    """Extract unique domain from an expression or BaseForm."""
+
     def __init__(
         self,
         compress: bool | None = True,
@@ -495,7 +497,7 @@ class UniqueDomainExtractor(DAGTraverser):
         from ufl.constant import Constant
         from ufl.geometry import GeometricQuantity
 
-        if isinstance(o, (Coefficient, Argument)):
+        if isinstance(o, Coefficient | Argument):
             fs = o.ufl_function_space()
             return fs.ufl_domain()
         elif isinstance(o, GeometricQuantity):
--- a/ufl/corealg/dag_traverser.py
+++ b/ufl/corealg/dag_traverser.py
@@ -1,8 +1,9 @@
 """Base class for dag traversers."""
 
 from __future__ import annotations
+
 from functools import singledispatchmethod, wraps
-from typing import overload, TYPE_CHECKING
+from typing import TYPE_CHECKING, overload
 
 if TYPE_CHECKING:
     from ufl.classes import Expr


ufl/domain.py
```diff
@@ -444,3 +430,136 @@ def find_geometric_dimension(expr):
...
     (gdim,) = gdims
     return gdim
+
+
+class UniqueDomainExtractor(DAGTraverser):
+    def __init__(
+        self,
+        compress: Union[bool, None] = True,
+        visited_cache: Union[dict[tuple, Expr], None] = None,
+        result_cache: Union[dict[Expr, Expr], None] = None,
```
leo-collins, line 440: Use `|` instead of `Union`

connorjward (commented):
I wonder if this could be generalised a little to a "domain extractor" that returns all domains. `extract_unique_domain` is then a special case of that.

ufl/domain.py
```diff
@@ -444,3 +430,145 @@ def find_geometric_dimension(expr):
...
+    def _(self, o: Expr, *operand_results) -> AbstractDomain:
+        """Process Indexed object by extracting the domain corresponding to the index."""
+        from ufl.functionspace import FunctionSpace
+
+        expression, multiindex = o.ufl_operands
+        expression_domain = operand_results[0]
+
+        if isinstance(expression_domain, MeshSequence):
+            index = multiindex[0]._value
+            element = expression.ufl_element()
```
ksagiyam, line 474: This assumes that `o.ufl_operands[0]` has `ufl_element()`, so something like the following will fail:
```
cell = triangle
elem0 = LagrangeElement(cell, 1)
elem1 = LagrangeElement(cell, 1)
elem = MixedElement([elem0, elem1])
mesh1 = Mesh(LagrangeElement(cell, 1, (2,)), ufl_id=100)
mesh2 = Mesh(LagrangeElement(cell, 1, (2,)), ufl_id=101)
domain = MeshSequence([mesh1, mesh2])
V = FunctionSpace(domain, elem)
f = Coefficient(V)
d = extract_unique_domain(
    Indexed(
        grad(f),
        MultiIndex((FixedIndex(0), FixedIndex(1))),
    )
)
```
It's probably hard to make `postorder`s robustly handle this sort of things.
First alternative that comes to my mind is `preorder` (or whatever you call it):
``` 
@process.register(Indexed)
def _(self, o, multiindex):
    expression, multiindex = o.ufl_operands
    return self(expression, multiindex=multiindex)

@process.register(Grad)
def _(self, o, multiindex):
    return self(o.ufl_operands[0], multiindex[:-1])

... and so on

```
You can see https://github.com/FEniCS/ufl/blob/main/ufl/algorithms/apply_coefficient_split.py.
`CoefficientSplitter` works as we make sure that `ReferenceValue`, `ReferenceGrad`, and `Restricted` have already been propagated to directly wrap terminals by the time we use it; you might need some "balancing" of this kind before calling `extract_unique_domain`, too, but it could be inconvenient.

Do you only need this for indexed `Argument`s and `Coefficient`s? What is the most general use case in your mind?

ufl/domain.py
```diff
@@ -444,3 +430,145 @@ def find_geometric_dimension(expr):
...
+
+    @process.register(Terminal)
+    @DAGTraverser.postorder
+    def _(self, o: Expr) -> AbstractDomain:
+        from ufl.argument import Argument
+        from ufl.coefficient import Coefficient
+        from ufl.constant import Constant
+        from ufl.geometry import GeometricQuantity
+
+        if isinstance(o, Coefficient | Argument):
```
ksagiyam, line 500: Should define separate singledispatch method as:
```
@process.register(Coefficient)
@process.register(Argument)
def _(self, o, ...):
    ...
```
Same for the other types.

test/test_unique_domain_extractor.py
```diff
@@ -0,0 +1,301 @@
...
+    cell = triangle
+    mesh1 = Mesh(LagrangeElement(cell, 1, (2,)), ufl_id=400)
+    mesh2 = Mesh(LagrangeElement(cell, 1, (2,)), ufl_id=401)
+    scalar_elem = LagrangeElement(cell, 1)
+
+    V1 = FunctionSpace(mesh1, scalar_elem)
+    V2 = FunctionSpace(mesh2, scalar_elem)
+
+    A = Matrix(V1, V2)
+    assert extract_unique_domain(A) == mesh1
```
ksagiyam, line 274: Is this what we expect?

ufl/domain.py
```diff
@@ -410,26 +416,6 @@ def extract_domains(expr: Expr | Form, expand_mesh_sequence: bool = True):
...
-    """
-    domains = extract_domains(expr, expand_mesh_sequence=expand_mesh_sequence)
-    if len(domains) == 1:
-        return domains[0]
-    elif domains:
-        raise ValueError("Found multiple domains, cannot return just one.")
-    else:
-        return None
-
-
```
ksagiyam, line 432: Can we not change the definition of `extract_unique_domain()` here? Instead, can we change
```
        for t in traverse_unique_terminals(expr):
            domainlist.extend(t.ufl_domains())
```
in `extract_domains()` function to use `DAGTraverser`? I think it is important to make sure that `extract_domains()` and `extract_unique_domains()` are consistent.

ksagiyam (changes requested):

Issue #415, opened 2025-09-04
Building the documentation produces many warnings and some errors


connorjward:
E.g. see [this workflow](https://github.com/FEniCS/ufl/actions/runs/17440496452/job/49521957607):
```
reading sources... [ 96%] releases/v2018.1.0
reading sources... [100%] releases/v2019.1.0
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/action.py:docstring of ufl.action.Action:5: WARNING: Definition list ends without a blank line; unexpected unindent. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/classes.py:docstring of ufl.action.Action:5: WARNING: Definition list ends without a blank line; unexpected unindent. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/classes.py:docstring of ufl.form.Form.terminal_numbering:5: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/classes.py:docstring of ufl.form.Form.terminal_numbering:5: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/domain.py:docstring of ufl.domain.MeshSequence:12: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/form.py:docstring of ufl.form.Form.terminal_numbering:5: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/form.py:docstring of ufl.form.Form.terminal_numbering:5: WARNING: Inline interpreted text or phrase reference start-string without end-string. [docutils]
/opt/hostedtoolcache/Python/3.13.7/x64/lib/python3.13/site-packages/ufl/formoperators.py:docstring of ufl.formoperators.adjoint:7: WARNING: Inline emphasis start-string without end-string. [docutils]
[...]

PR #421, opened 2025-09-10
Fix documentation errors

schnellerhase:
Addresses partially https://github.com/FEniCS/ufl/issues/415.

Warning resolution not possible due to https://github.com/sphinx-doc/sphinx/issues/11225.


connorjward:
> I checked the sphinx options, but only found options to error on warning and alike. To be honest I am somewhat surprised that an 'error' results in a passing `make` call.

Yeah that is quite odd.

PR #422, opened 2025-09-19
Check `Interpolate` argument numbering

leo-collins:
Adds some checks on the argument numberings to the `Interpolate` form.


jorgensd:
To me this looks good. However, as this mainly touches firedrake specific functionality (for now), I'd like @dham opinion.

PR #423, opened 2025-09-29
Expose `ufl.Label`

jorgensd:
Currently unreachable as it is within `ufl.variable`, which is overloaded by the `variable` function at import

### DIFF
--- a/ufl/__init__.py
+++ b/ufl/__init__.py
@@ -439,6 +439,7 @@ from ufl.tensors import (
     unit_vectors,
 )
 from ufl.utils.sequences import product
+from ufl.variable import Label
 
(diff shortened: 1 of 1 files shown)


test/test_derivative.py
```diff
@@ -951,3 +953,34 @@ def NS_a(u, v):
     L = NS_a(U, v) * dx
     _ = derivative(L, U, du)
     # TODO: assert something
+
+
+def test_variable_label():
+    """Check that deriviatives respect variable labels"""
```
jorgensd, line 959:
```suggestion
    """Check that derivatives respect variable labels"""
```

PR #426, opened 2025-10-03
`extract_arguments` on nested `BaseFormOperator`

leo-collins:
Fix `extract_arguments` to work with nested base form operators. Needed for https://github.com/FEniCS/ufl/pull/425

### DIFF
--- a/ufl/algorithms/analysis.py
+++ b/ufl/algorithms/analysis.py
@@ -106,7 +106,7 @@ def extract_type(a, ufl_types):
             # Filter out any BaseFormOperators if necessary
             # from the recursively extracted results
             if remove_base_form_ops:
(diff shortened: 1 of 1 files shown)


leo-collins:
If we did

```
u1 = Cofunction(V1.dual())
u2 = Interpolate(Argument(V2, 0), u1)
u3 = Interpolate(Argument(V3, 0), u2)
args = extract_type(u3, BaseArgument)
```

then `args` contained an `Interpolate` which is wrong

pbrubeck:
If we have a primal `Expr` involving `BaseFormOperators`, the way to identify uncontracted arguments should be to cast into a `BaseForm` and then call `BaseForm.arguments()`.

leo-collins:
In this case though `u3.arguments()` only contains `Argument(V3, 0)`

PR #434, opened 2025-10-21
Try splitting python-app workflow into separable steps

jorgensd:
The monolithic `pythonapp.yml` is broken into separable steps, which are all called from `deploy.yml`, making it easier to pinpoint issues, and run manual triggers.

Logic is now:

1. Lint code
2. If 1 is successful run test matrix on python and os version
   2b) if successful, upload coveralls if on main or on a tag.
3. If 2 is successful build and upload documentation.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,33 +2,32 @@ name: Build UFL Documentation
 on:
   workflow_call:
     inputs:
-        deploy:
-            description: 'Deploy documentation'
-            required: false
-            default: false
-            type: boolean
+      deploy:
+        description: "Deploy documentation"
+        required: false
+        default: false
+        type: boolean
 
   workflow_dispatch:
     inputs:
-        deploy:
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,68 @@
+name: Build UFL Documentation
```
jhale: The deployment of the documentation should be in this file, and only run if a parameter deploy is true.

.github/workflows/pythonapp.yml
```diff
@@ -1,99 +1,93 @@
...
+        deploy:
+            description: 'Whether to deploy documentation'
+            required: false
+            default: false
+            type: boolean
+
+  workflow_dispatch:
+    inputs:
+        python-version:
+          description: 'Python version to use'
```
jorgensd:
```suggestion
          description: 'Python version'
```

PR #447, opened 2025-11-24
Fix action interpolate simplification

leo-collins:
The simplification `Action(OneForm, Interpolate(Expr, Coargument)) -> Interpolate(Expr, OneForm)` was failing in the case where `Expr` had an argument due to the new numbering checks. I think this worked before because in Firedrake we were renumbering arguments inside `Interpolate`. This was pointed out here: https://github.com/firedrakeproject/firedrake/discussions/4597

### DIFF
--- a/test/test_interpolate.py
+++ b/test/test_interpolate.py
@@ -35,7 +35,7 @@ from ufl.algorithms.analysis import (
 )
 from ufl.algorithms.expand_indices import expand_indices
 from ufl.core.interpolate import Interpolate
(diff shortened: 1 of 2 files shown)


leo-collins:
Firedrake tests passing here: https://github.com/firedrakeproject/firedrake/pull/4751

dham (dismissed):
This looks exactly right to me, thanks also for the tests.

dham (approved):

PR #448, opened 2025-11-27
Fix split(SymmetricTensor)

pbrubeck:
Fixes nested `split()` on mixed spaces with symmetric tensor-valued `Coefficient`/`Argument`.

Fixes https://github.com/firedrakeproject/firedrake/issues/4752

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build sdist and wheel
         run: python -m build .
 
-      - uses: actions/upload-artifact@v5
+      - uses: actions/upload-artifact@v6
         with:
           path: dist/*
 
@@ -66,7 +66,7 @@ jobs:
       id-token: write
 
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: artifact
           path: dist
@@ -88,7 +88,7 @@ jobs:
       id-token: write
 
     steps:
-      - uses: actions/download-artifact@v6
+      - uses: actions/download-artifact@v7
         with:
           name: artifact
           path: dist
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -40,7 +40,7 @@ jobs:
           python -m sphinx -W -b html source/ build/html/
 
       - name: Upload documentation artifact
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: documentation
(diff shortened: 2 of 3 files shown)


ufl/split_functions.py
```diff
@@ -33,10 +36,11 @@ def split(v):
 
     elif isinstance(v, ListTensor):
         # Special case: split previous output of split again
-        ops = v.ufl_operands
+        ops = tuple(v[i] for i in np.ndindex(v.ufl_shape))
+
         if all(isinstance(comp, Indexed) for comp in ops):
             args = [comp.ufl_operands[0] for comp in ops]
-            if all(args[0] == args[i] for i in range(1, len(args))):
+            if len(args) == product(v.ufl_shape) and all(args[0] == a for a in args[1:]):
```
pbrubeck:
```suggestion
            if all(args[0] == args[i] for i in range(1, len(args))):
```

ufl/split_functions.py
```diff
@@ -85,8 +89,18 @@ def split(v):
...
+    if isinstance(element.pullback, SymmetricPullback):
+        symmetry = element.pullback._symmetry
+        indices = (comp for idx, comp in sorted(symmetry.items()))
+    else:
+        indices = range(len(element.sub_elements))
+
+    domains = tuple(domain.iterable_like(element))
+    for i in indices:
+        d = domains[i]
+        e = element.sub_elements[i]
```
pbrubeck: Should `SymmetricElement.sub_elements` be returning the tuple with the symmetries applied or only the linearly independent ones?

ufl/split_functions.py
```diff
@@ -7,11 +7,14 @@
 #
 # Modified by Anders Logg, 2008
```
jorgensd: @pbrubeck add your name here?:)

pbrubeck: done

pbrubeck:
> Note that it works weil with `ufl.MixedFunctionSpace`.
> 
> I can confirm that this branch fixes the issue.

The `ufl.MixedFunctionSpace` approach does not create a `Coefficient(Z)`, but two separate ones in `V` and `S`. Is that what you intended?

pbrubeck:
> We usually target main, and then backport to release, rather than the opposite. @jhale can comment on this

Should I target main instead?

pbrubeck:
@jorgensd I changed the target to main. Is there anything else to be done to get this PR merged?

pbrubeck:
@jorgensd could you have a look at the dolfinx failure in CI?

PR #451, opened 2025-12-09
replace: only expand CoefficientDerivative

pbrubeck:
`replace` was unecessarily expanding all derivatives too early.  This becomes a problem in Irksome where we introduce a `TimeDerivative` that cannot be expanded in UFL. 

This PR avoids this problem by expanding `CoefficientDerivative` only.

### DIFF
--- a/ufl/algorithms/apply_derivatives.py
+++ b/ufl/algorithms/apply_derivatives.py
@@ -2292,3 +2292,90 @@ def apply_coordinate_derivatives(expression):
     """Apply coordinate derivatives to an expression."""
     dag_traverser = CoordinateDerivativeRuleDispatcher()
     return map_integrands(dag_traverser, expression)
+
+
+class CoefficientDerivativeRuleDispatcher(DAGTraverser):
+    """Dispatcher."""
+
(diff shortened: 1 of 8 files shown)


jorgensd (approved):
LGTM

ufl/algorithms/apply_derivatives.py
```diff
@@ -2292,3 +2295,62 @@ def apply_coordinate_derivatives(expression):
...
+        full pipeline, not the narrowed, CoefficientDerivative-only
+        treatment `self` gives to everything *outside* this node.
+        """
+        expr, w, v, cd = o.ufl_operands
+        full_dispatcher = self._dag_traverser_cache.setdefault(
+            (DerivativeRuleDispatcher,),  # type: ignore
+            DerivativeRuleDispatcher(),  # type: ignore
+        )
+        f = full_dispatcher(expr)  # type: ignore
+        key = (GateauxDerivativeRuleset, w, v, cd)
```
pbrubeck, line 2339: This defeats the purpose of this DAGTraverser, I think.

PR #470, opened 2026-03-19
Add `arity` property to form

schnellerhase:
There is currently no better way than `len(form.arguments())` to check for the arity/rank of a form. Introduces property for simpler access. Since `rank` is already associated with tensor ranks in ufl, use the name `arity`.

If a form contains an ambiguous arity, i.e. terms with different argument counts `None` is returned.

Ref https://github.com/FEniCS/dolfinx/pull/4118.

### DIFF
--- a/ufl/form.py
+++ b/ufl/form.py
@@ -340,6 +340,11 @@ class Form(BaseForm):
         """Returns whether the form has no integrals."""
         return len(self.integrals()) == 0
 
+    @property
+    def arity(self):
+        """Arity of the form."""
+        return len(self.arguments())
+
     def ufl_domains(self):
         """Return the geometric integration domains occuring in the form.
 
--- a/ufl/form.py
+++ b/ufl/form.py
@@ -341,9 +341,28 @@ class Form(BaseForm):
         return len(self.integrals()) == 0
 
     @property
-    def arity(self):
-        """Arity of the form."""
-        return len(self.arguments())
+    def arity(self) -> int | None:
+        """Arity of the form.
+
+        Returns:
+            Number of arguments if all integrals share the argument count, otherwise None.
+        """
+        if not self._integrals:
+            return 0
+
+        from ufl.algorithms.analysis import extract_terminals_with_domain
+
+        arity = None
+        for integral in self._integrals:
+            args, _, _ = extract_terminals_with_domain(integral.integrand())
+            _arity = len(set(args))
+
+            if arity is None:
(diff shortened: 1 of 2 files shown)


schnellerhase:
I changed over to return `int | None`. If the arity is the same for all integrals it is returned, if we have different arities it returns `None`. This allows to differentiate `u  * v * dx` and `u * v * dx + v * dx`- which would with the max have both returned 2.

ufl/form.py
```diff
@@ -340,6 +340,37 @@ def empty(self):
...
+        Note:
+            Mixed function spaces are supported if the parts are not mixed in a single integral.
+
+        Returns:
+            Number of arguments if all integrals share the argument count, otherwise None.
+        """
+        if not self._integrals:
+            return 0
+
+        from ufl.algorithms.analysis import extract_terminals_with_domain
```
jorgensd: Do we need this here to avoid circular imports?

ufl/form.py
```diff
@@ -340,6 +340,37 @@ def empty(self):
...
+
+        arity = None
+        for integral in self._integrals:
+            args, _, _ = extract_terminals_with_domain(integral.integrand())
```
jorgensd: Why not just use `extract_arguments` (https://github.com/FEniCS/ufl/blob/0e7e2eedeb0cf9c18c794aa0df7bd62278067a07/ufl/algorithms/analysis.py#L172-L178)?

ufl/form.py
```diff
@@ -340,6 +340,37 @@ def empty(self):
...
+
+            if len(set(arg.part() for arg in args)) > 1:
+                raise RuntimeError("Arity does not support mixed arguments in an integral.")
```
jorgensd, line 363: What case does this cover?

jorgensd (changes requested):

ufl/form.py
schnellerhase: Yes exactly. Also used in other parts of `form.py` in the same manner.

ufl/form.py
schnellerhase: Works around the problems arising from having mixed arguments present in the form https://github.com/FEniCS/ufl/pull/470/changes#diff-40318d7ac6b526c958053e25f90768d68e58d7c26e6479fe580f7420f53de86fR146-R159. Probably this is too restrictive. But I am not quite sure what the right checks should be here instead.

PR #481, opened 2026-04-24
Use DAGTraverser instead of MultiFunction for ufl2unicode

connorjward:


dham (changes requested):
More than no tests please, then this is good.

dham (approved):

Issue #268, opened 2024-04-18
Form should have a quadrature rule


mscroggs:
The metadata of a form is commonly used to store a quadrature rule; it would be better support giving a quadrature rule to a form. To do this we need to:

- [ ] Add `AbstractBaseQuadrature` class
- [ ] Add `DefaultQuadrature` subclass of this
- [ ] Add `quadrature_rule` property to `Form`
- [ ] Optionally allow `Form` to take a quadrature rule when it is created (with default value `DefaultQuadrature`)

PR #483, opened 2026-05-09
Enhance Measure.__call__

pbrubeck:
This PR adds kwargs to `Measure.__call__` to support `dx(intersect_mesures=...)`.

These changes are not API-breaking, as we appends kwargs at the end on the list.

### DIFF
--- a/ufl/measure.py
+++ b/ufl/measure.py
@@ -287,7 +287,7 @@ class Measure:
         subdomain_data=None,
         degree=None,
         scheme=None,
-        max_quadrature_degree=None,
+        max_degree=None,
         intersect_measures=None,
     ):
         """Reconfigure measure with new domain specification or metadata."""
@@ -311,14 +311,14 @@ class Measure:
         # If degree or scheme is set, inject into metadata. This is a
         # quick fix to enable the dx(..., degree=3) notation.
         # TODO: Make degree and scheme properties of integrals instead of adding to metadata.
-        if (degree, scheme, max_quadrature_degree) != (None, None, None):
+        if (degree, scheme, max_degree) != (None, None, None):
             metadata = {} if metadata is None else metadata.copy()
             if degree is not None:
                 metadata["quadrature_degree"] = degree
             if scheme is not None:
                 metadata["quadrature_rule"] = scheme
-            if max_quadrature_degree is not None:
-                metadata["max_quadrature_degree"] = max_quadrature_degree
+            if max_degree is not None:
+                metadata["max_quadrature_degree"] = max_degree
 
         # If we get any keywords, use them to reconstruct Measure.
         # Note that if only one argument is given, it is the
--- a/ufl/measure.py
+++ b/ufl/measure.py
@@ -287,7 +287,6 @@ class Measure:
         subdomain_data=None,
         degree=None,
         scheme=None,
-        max_degree=None,
         intersect_measures=None,
     ):
         """Reconfigure measure with new domain specification or metadata."""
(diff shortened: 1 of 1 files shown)


ufl/measure.py
```diff
@@ -287,6 +287,8 @@ def __call__(
         subdomain_data=None,
         degree=None,
         scheme=None,
+        max_quadrature_degree=None,
```
pbrubeck:
```suggestion
        max_degree=None,
```

ufl/measure.py
```diff
@@ -309,12 +311,14 @@ def __call__(
         # If degree or scheme is set, inject into metadata. This is a
         # quick fix to enable the dx(..., degree=3) notation.
         # TODO: Make degree and scheme properties of integrals instead of adding to metadata.
-        if (degree, scheme) != (None, None):
+        if (degree, scheme, max_quadrature_degree) != (None, None, None):
```
pbrubeck:
```suggestion
        if (degree, scheme, max_degree) != (None, None, None):
```

ufl/measure.py
```diff
@@ -309,12 +311,14 @@ def __call__(
...
             metadata = {} if metadata is None else metadata.copy()
             if degree is not None:
                 metadata["quadrature_degree"] = degree
             if scheme is not None:
                 metadata["quadrature_rule"] = scheme
+            if max_quadrature_degree is not None:
+                metadata["max_quadrature_degree"] = max_quadrature_degree
```
pbrubeck:
```suggestion
            if max_degree is not None:
                metadata["max_quadrature_degree"] = max_degree
```

ufl/measure.py
```diff
@@ -287,6 +287,8 @@ def __call__(
         subdomain_data=None,
         degree=None,
         scheme=None,
+        max_degree=None,
```
jorgensd: As `max_degree` for now is part of metadata, it should strictly not be exposed in the call.
Instead, it should be passed through the metadata in reconstruct:
```python
dx = Measure("dx", domain=mesh)
dx_max = dx(metadata={"max_degree": 5})
```

ufl/measure.py
```diff
@@ -287,6 +287,8 @@ def __call__(
...
+        intersect_measures=None,
```
jorgensd, line 290: This is fine, as it more closely resembles the `Measure` constructor.

jorgensd (changes requested):

ufl/measure.py
pbrubeck: Currently `dx(metadata={"max_quadrature_degree": 5})` works. The `max_degree` kwarg is added here to be consistent with `dx(degree=5)`. 

We originally had to explain the inconsistent shorthand in the [Firedrake manual](https://www.firedrakeproject.org/quadrature.html#id1).

jorgensd: I'd rather go the opposite way, removing `degree` as reconstruction and initialiser argument, as this is currently a compiler instruction, and doesn't relate to the form language. @mscroggs and @dham talked about formalizing quadrature-rule a while back: https://github.com/FEniCS/ufl/issues/268

pbrubeck: I think we should keep this call interface as user-friendly as possible. I don't think we make anyone happy by breaking codes that already worked. 

The metadata interface is quite ugly for users. How about automatically adding any unrecognized kwargs (kwargs that do not relate to the form language) to the metadata?

ufl/measure.py
```diff
@@ -309,12 +311,14 @@ def __call__(
         # If degree or scheme is set, inject into metadata. This is a
         # quick fix to enable the dx(..., degree=3) notation.
         # TODO: Make degree and scheme properties of integrals instead of adding to metadata.
-        if (degree, scheme) != (None, None):
+        if (degree, scheme, max_degree) != (None, None, None):
```
pbrubeck:
```suggestion
        if (degree, scheme) != (None, None):
```

ufl/measure.py
```diff
@@ -309,12 +311,14 @@ def __call__(
...
             metadata = {} if metadata is None else metadata.copy()
             if degree is not None:
                 metadata["quadrature_degree"] = degree
             if scheme is not None:
                 metadata["quadrature_rule"] = scheme
+            if max_degree is not None:
+                metadata["max_quadrature_degree"] = max_degree
```
pbrubeck:
```suggestion
```

ufl/measure.py
pbrubeck:
```suggestion
```

pbrubeck: Well, perhaps another day.

jorgensd (approved):

PR #485, opened 2026-05-21
Drop Python 3.10 support

jhale:
Python 3.10 moves out of security updates at the end of summer.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ description = "Unified Form Language"
 readme = "README.md"
 license = "LGPL-3.0-or-later"
 license-files = ["COPYING*"]
-requires-python = ">=3.10.0"
+requires-python = ">=3.11"
 dependencies = ["numpy"]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -17,7 +17,7 @@ description = "Unified Form Language"
 readme = "README.md"
 license = "LGPL-3.0-or-later"
 license-files = ["COPYING*"]
-requires-python = ">=3.10.0"
+requires-python = ">=3.11"
```
connorjward, line 20: I understand the desire to stop testing 3.10 but does this line really need changing? You aren't introducing 3.10 incompatible changes here.

jhale: Well, if it's not tested, it "doesn't work" :)

That aside, my view is it's better to signal modernised language/library support to developers.

connorjward: I totally understand your point of view. I just don't want to cause unnecessary churn for users.

Ultimately I don't really mind. If you feel like dropping 3.10 then that's no problem.

And now that we have our joint release strategy this won't hit our users until October when 3.10 is definitely end of life.

connorjward (approved):

mscroggs (approved):
Finally, we can do https://github.com/FEniCS/ufl/issues/176

PR #493, opened 2026-06-24
Support string subdomain IDs

pbrubeck:


jorgensd:
Is there a particular use-case you have in mind for string domain ids?
In general, I think we should revisit the whole `ufl.Measure` implementation, as it seems a bit limiting.
Especially the integral_types being strings rather than a typing Protocol.
Whenever we add strings we add tons of undefined behavior (or undocumented behavior).

pbrubeck:
> Is there a particular use-case you have in mind for string domain ids?
> In general, I think we should revisit the whole `ufl.Measure` implementation, as it seems a bit limiting.
> Especially the integral_types being strings rather than a typing Protocol.
> Whenever we add strings we add tons of undefined behavior (or undocumented behavior).


I can update docstrings. I'm fine if this is revisited. We should avoid breaking API.

https://github.com/firedrakeproject/firedrake/pull/5199

jorgensd:
I guess this is ok to do, as very little of the functionality with UFL relies on a specific typing literal.

jorgensd (approved):

PR #494, opened 2026-07-11
Fix ZeroBaseForm holes in derivative expansion and action

pbrubeck:
## Summary
- `GateauxDerivativeRuleset` had no rule for `ZeroBaseForm` itself, so differentiating an already-zero `BaseForm` (e.g. a Hessian of a `Matrix`) raised `AssertionError: Rule not set for ZeroBaseForm` in `expand_derivatives`.
- `map_integrands`'s `FormSum` branch collapsed to a bare `0` via `sum([])` when every component vanished, silently dropping the argument shape; `apply_derivatives` had a matching fallback that rebuilt an argument-less `Form([])` instead of a properly shaped `ZeroBaseForm`.
- `action()` only expanded derivatives when the top-level object was already a `Form`/`BaseFormOperator`, so a `CoefficientDerivative` reachable only through a `FormSum` (e.g. from an affine residual `F - b`) survived unexpanded inside the resulting lazy `Action`. Widened the check to match `adjoint()`'s existing `BaseForm` handling.
- `ZeroBaseForm` inherited `BaseForm.empty()`, which always returns `False` — gave it its own `empty()` returning `True`.

## Test plan
- [x] `pytest test/` (full suite, 969 passed)
- [x] New regression tests added to `test/test_duals.py`: nested differentiation of a `Matrix` (Hessian), a `FormSum` of `Cofunction`s whose components all vanish under differentiation, and `action()` on a `FormSum` no longer leaving a `CoefficientDerivative` behind.
- [x] `ruff check`


pbrubeck:
Note from Claude refering to https://github.com/firedrakeproject/firedrake/blob/49f19c307b1aa1cfa1386705f31ed72975e65c23/firedrake/formmanipulation.py#L92

I did not generalize this to `map_integrands's` plain-Form branch — I tried making it return `ZeroBaseForm` there too (which would have let formmanipulation.py's TODO disappear entirely), but it's provably wrong for transforms like ExtractSubBlock that replace arguments' function spaces rather than just adding a new one: a live regression (`test_slate_infrastructure.py::test_blocks`) showed it silently returning arguments on the wrong (unsplit) space. I reverted that piece. So the `formmanipulation.py` TODO and the identical pre-existing workaround in `matrix_free/operators.py:150` stay as manual reconstructions — but now backed by an accurate comment explaining why expand_derivatives is genuinely required (the zero cancellation is a chain-rule consequence, not something detectable without differentiating), rather than the old "avoid expand_derivatives" aspiration.

PR #495, opened 2026-07-11
Add AGENTS.md and CLAUDE.md

pbrubeck:
## Summary
- Adds `AGENTS.md`, documenting UFL's architecture (the `Expr`/`BaseForm` primal-dual split, the `singledispatchmethod`-based `DAGTraverser` dispatch pattern in `apply_derivatives.py`, `map_integrands`'s recursive form transform), coding conventions (Google-style docstrings, the `# type: ignore` convention on AD rule registrations, ruff's 100-char line length), testing conventions, and the fork-push/origin-PR + three-stage-lint workflow this repo uses.
- The Anti-Patterns section is written directly from real bugs: using a bare `0`/`Form([])` instead of `ZeroBaseForm` for a form that turned out to be zero (losing argument shape) and a documented dead end from that same investigation (#494 — naively generalizing `map_integrands`'s `Form` branch the way its `FormSum` branch already works), plus a third worked example from fixing `pbrubeck/fix/replace-coefficient-derivative`'s `CoefficientDerivativeRuleDispatcher`: narrowing a dispatcher's node-type coverage (to protect foreign `Derivative` subtypes) without narrowing *where* that narrowing applies, which silently also left spatial `Grad` un-normalized inside a `CoefficientDerivative`'s own content.
- Adds a **Pattern Matching For Planning And Debugging** section (moved to just before Anti-Patterns) drawing on #494, #496 (Jacobian cancellation), and the `CoefficientDerivativeRuleDispatcher` fix: search sibling/downstream packages for prior art before designing a new algorithm (#496's `JacobianCanceller`/`IdentityEliminator` deliberately ports GEM's own Delta-cancellation passes), match a new dispatch rule to the nearest *structurally* similar existing one, classify which structural category a fix belongs to before generalizing it to a sibling code path, isolate cross-package regressions with a one-delta reproduction, and — new — treat a fix that trades one failing test for another as a context signal (type-keyed dispatch can't distinguish "inside a region" from "outside it") rather than a cue to add more special cases.
- The Debugging section gained two entries from the same fix: `DAGTraverser` tracebacks are mostly `KeyError` cache-miss noise (skip to the last exception, read `self`'s class at each frame to see which dispatcher/ruleset was active), and diffing a narrow code path against the full pipeline on the same input localizes exactly which node differs.
- Adds `CLAUDE.md` importing `AGENTS.md` so Claude Code picks it up automatically, mirroring the pattern already in use downstream in [Firedrake](https://github.com/firedrakeproject/firedrake/blob/main/CLAUDE.md).

## Test plan
- [x] N/A (documentation only)


pbrubeck:
Still iterating on this

AGENTS.md
```diff
@@ -0,0 +1,336 @@
...
+  to `FEniCS/ufl:main`:
+  ```bash
+  git checkout -b <branch> main   # branch off an up-to-date origin/main
+  git push -u fork <branch>
+  gh pr create --repo FEniCS/ufl --base main --head <fork-org>:<branch> --title ... --body ...
+  ```
+  A permission error pushing to `origin` is expected, not a sign of misconfiguration.
+* Run all three `lint.yml` stages before pushing — `ruff check .` passing alone is not enough:
+  ```bash
+  ruff check . && ruff format --check . && mypy -p ufl && mypy test/ && (cd test && mypy ../demo/)
```
pbrubeck, line 88: We should be taking this as an invitation to write a Makefile with `make lint`

PR #496, opened 2026-07-11
Cancel contractions of the Jacobian with its inverse

pbrubeck:
## Description

Piola-mapped elements generate expressions of the form `J * inv(J)` that were not being simplified, because `JacobianInverse` was expanded into individual matrix entries too early in the form preprocessing.

This PR delays that expansion: during geometry lowering, `Jacobian`, `JacobianInverse`, and `JacobianDeterminant` are preserved as opaque terminals (which also lets `is_cellwise_constant` fold their derivatives to zero early on affine cells). A new algorithm module `cancel_jacobian_products` then simplifies the integrand in three traversals:

1. `JacobianCanceller`: `IndexSum(J[a, k] * K[k, b] * factors, k) -> Identity[a, b] * factors` (and the K·J order, which is also valid for pseudo-inverses on immersed manifolds). Contractions are chased through nested `IndexSum`s by interchanging the order of summation.
2. `IdentityEliminator`: `IndexSum(Identity[a, k] * factors, k) -> factors[k -> a]`, and folds fixed-index Identity entries into scalar constants. This also cleans up user-written Identity contractions.
3. `ReciprocalCanceller`: cancels reciprocal factors within products, such as `detJ**2 * (1/detJ)**2 -> 1`.

The surviving Jacobian quantities are lowered as usual afterwards, so form compilers see no new terminal types. The behaviour is opt-in via a new `compute_form_data` kwarg `do_cancel_jacobian_products` (default off, so ffcx is unaffected).

For the Johnson-Mercier element, the div-div integrand shrinks from 9630 to 664 characters, becoming literally `(1/detJ**2) J : rdiv(sigma)` on affine cells. Together with companion GEM changes this gives up to -42% flops and -51% compile time (see companion PRs in firedrakeproject/fiat and firedrakeproject/firedrake).


test/test_cancel_jacobian_products.py
```diff
@@ -0,0 +1,105 @@
...
+
+    integrand_plain = fd_plain.preprocessed_form.integrals()[0].integrand()
+    integrand_cancelled = fd_cancelled.preprocessed_form.integrals()[0].integrand()
+
+    # The whole point of the pass: the div-div integrand of a
+    # contravariant Piola-mapped element is dominated by J-Jinv
+    # contractions that cancel, so the cancelled integrand is much
+    # smaller than the one form compilers would otherwise have to deal
+    # with, and does not introduce any new terminal type of its own.
+    assert len(str(integrand_cancelled)) < len(str(integrand_plain)) / 4
```
pbrubeck: Assert that Jacobians arent there with extract_type. Also compare against hand-coded expected result for mathematical correctness.

ufl/algorithms/cancel_jacobian_products.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def match(self, with_k, rest, k):
+        """Simplify IndexSum(product(with_k + rest), k), or return None.
+
+        Args:
+            with_k: factors that have k as a free index.
+            rest: factors that do not depend on k.
+            k: the summation index.
+        """
+        raise NotImplementedError(f"match() not implemented by {type(self).__name__}.")
```
jorgensd: Use `@abstractmethod` instead to make it clear that this has to be implemented for any sub-class?

pbrubeck: Done. Firedrake CI is not passing at the moment: zany elements that are H1 but Piola mapped are failing. So I'll make sure it passes before we merge.

pbrubeck: We needed this fix https://github.com/firedrakeproject/fiat/pull/255, which should go in before the UFL change

jorgensd (approved):

PR #497, opened 2026-07-16
Fix Interpolate holes

### LINKED ISSUE #473: `compute_form_signature` is not correct for interpolate
Given the two different forms:
```python
import ufl
import basix.ufl

el = basix.ufl.element("N1curl", "triangle", 1)

el2 = basix.ufl.element("Lagrange", "triangle", 1, shape=(2,))

domain  = ufl.Mesh(basix.ufl.element("Lagrange", "triangle", 1, shape=(2,)))

x = ufl.SpatialCoordinate(domain)
V = ufl.FunctionSpace(domain, el)
V2 = ufl.FunctionSpace(domain, el2)

f = x

z =  ufl.Interpolate(f, V)[0]*ufl.dx
z2 =  ufl.Interpolate(f, V2)[0] * ufl.dx
print(z.signature())
print(z2.signature())
```
we get the same signature, as the space is not an UFL operand, thus:
```python

def compute_expression_hashdata(expression, terminal_hashdata) -> bytes:
    """Compute expression hashdata."""
    cache: dict[UFLObject, bytes] = {}

    for expr in unique_post_traversal(expression):
        # Uniquely traverse tree and hash each node
        # E.g. (a + b*c) is hashed as hash([+, hash(a), hash([*, hash(b), hash(c)])])
        # Traversal uses post pattern, so children hashes are cached
        if expr._ufl_is_terminal_:
            data = [terminal_hashdata[expr]]
        else:
            data = [expr._ufl_typecode_]

            for op in expr.ufl_operands:
                data += [cache[op]]
        cache[expr] = hashlib.sha512(str(data).encode("utf-8")).digest()
    return cache[expression]

```
will yield the same hash-data for both of them. @dham, @mscroggs how should we deal with this?

pbrubeck:
## Promote ufl.Interpolate to first-class citizen

Adds preprocessing needed for TSFC kernel generation for a `Form` with nested `Interpolate` nodes or plain `Interpolate`.

- Fixes `ufl_shape` and `__neg__` for primal `Interpolate`
- Add support for terminal modifiers (`Restriction`, `ReferenceValue`, `ReferenceGradient`)
- Add degree estimation

Fixes https://github.com/FEniCS/ufl/issues/473


jorgensd (approved):
As support for ufl.interpolate is not supported in DOLFINx (yet https://github.com/FEniCS/ffcx/pull/829), I don't see any issue with this PR. 
Happy to merge whenever CI passes.

ufl/core/interpolate.py
```diff
@@ -67,12 +74,121 @@ def __init__(self, expr, v):
...
+
+            numbering = {expression: i for i, expression in enumerate(self.arguments())}
+            numbering.update({expression: i for i, expression in enumerate(self.coefficients())})
+            for expressions in exprs_by_type.values():
+                for i, expression in enumerate(sorted_by_count(expressions)):
+                    numbering.setdefault(expression, i)
+            self._terminal_numbering = numbering
+        return self._terminal_numbering
+
+    def signature(self):
```
jorgensd, line 120: Does this resolve: https://github.com/FEniCS/ufl/issues/473
If so can we add a test? @pbrubeck

pbrubeck: Done.

jorgensd (approved):

Issue #486, opened 2026-05-28
Function to extract linear combination if present in ufl_expression


jorgensd:
Typically, one would like to assign a linear combination of functions in the same function space to each other (for instance during time-stepping).
I encountered this when trying to get a DOLFINx Irksome implementation up and running:
https://github.com/firedrakeproject/Irksome/pull/234/
and in Legacy DOLFIN there is a rather large complicated section on this extraction:
https://bitbucket.org/fenics-project/dolfin/src/a8dcf753d3ab9ae19d7bedf8c634a4fe545032d2/python/dolfin/function/function.py?at=master#lines-37:177

Instead of repeating the same logic over and over in different implementations (legacy Dolfin, DOLFINx, firedrake) it would be nice to have this function native in UFL.

michalhabera:
See also https://github.com/fenics-dolfiny/dolfiny/blob/d2456e7d64c6de15384e219fbd0f8b1b0a81b544/src/dolfiny/expression.py#L143-L201 for a simple version of this. Useful also in a time-stepping context.

connorjward:
For a less simple version see https://github.com/firedrakeproject/firedrake/blob/main/firedrake/assign.py#L35. Something in UFL would be nice.

jorgensd:
So I guess what we want is a combination of the two, preferably using the DagTraverser?

connorjward:
> So I guess what we want is a combination of the two, preferably using the DagTraverser?

Yeah I think so. Note that our implementation is made more complicated by the existence of the Real space. In Firedrake a `Function` in R is the same thing as your UFL `Constant`. So in Firedrake we allow the user to treat a `Function` in R as the scalar in a linear combination. Ideally the implementation here will be extensible (via another `@process.register`?) so we can add a custom handler for that inside Firedrake.

We also want to allow the multiplication of vector-valued functions by vector-valued constants.

jorgensd:
> > So I guess what we want is a combination of the two, preferably using the DagTraverser?
> 
> Yeah I think so. Note that our implementation is made more complicated by the existence of the Real space. In Firedrake a `Function` in R is the same thing as your UFL `Constant`. So in Firedrake we allow the user to treat a `Function` in R as the scalar in a linear combination. Ideally the implementation here will be extensible (via another `@process.register`?) so we can add a custom handler for that inside Firedrake.
> 
> We also want to allow the multiplication of vector-valued functions by vector-valued constants.

[...]

PR #504, opened 2026-08-20
Extract linear combination from `ufl`-expression

jorgensd:
It is often common to extract linear combinations $\sum_i c_i u_i$ or a set of scalar weights $c_i$ and function $u_i\in V$ for quick assignment without having to call interpolation or projection, ref #486.

## AI assistance
The code and documentation was generated by prompting Gemini (August 2026) based on my own initial implementations without a DAGTraverser. 
I reviewed, tested, and take responsibility for the final contribution.

## Deprecated example after discussions in this PR
Initially to catch scalar valued real elements, one would have to overload the dag-traverser. This is no longer true, as we use `ufl.checks.is_scalar_constant_expression` to have a common base for both Firedrake and FEniCS.

The old example can be found below:
```python
import ufl
from ufl.algorithms.extract_linear_combination import LinearCombinationExtractor
from functools import singledispatchmethod

class CustomLinearCombinationExtractor(LinearCombinationExtractor):
    """Custom extractor to handle linear combinations of UFL expressions."""

    @singledispatchmethod
    def process(self, o: ufl.classes.Expr, **kwargs):
        """Fallback for any unsupported node types."""
        return super().process(o, **kwargs)

    @process.register(dolfinx.fem.Function)
    def _(self, o, **kwargs):
        if o.function_space.ufl_element().is_real and o.ufl_shape == ():
            return o
        return [(1.0, o)]


def extract_linear_combination(
    expr: ufl.core.expr.Expr,
) -> list[tuple[ufl.core.expr.Expr, ufl.classes.Coefficient]]:
    """Wrapper to initialize traverser and extract linear combinations.

    Returns:
        A list of tuples where the first element is the UFL expression of the
        weight, and the second element is the base UFL Coefficient (spatial function).
    """
    extractor = CustomLinearCombinationExtractor()
    final_result = extractor(expr)

    if not isinstance(final_result, list):
        raise ValueError("Expression evaluated to a pure scalar, no spatial functions found.")

    return final_result
```

### DIFF
--- a/test/test_extract_linear_combination.py
+++ b/test/test_extract_linear_combination.py
@@ -1,3 +1,9 @@
+# Copyright (C) 2026 Jørgen S. Dokken
+#
+# This file is part of UFL (https://www.fenicsproject.org)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+
 import pytest
 from utils import LagrangeElement
 
--- a/test/test_extract_linear_combination.py
+++ b/test/test_extract_linear_combination.py
@@ -139,3 +139,46 @@ def test_negative(V, domain):
 
     assert (-r_func, u) in res
     assert (2.0, u) in res
+
+
+def test_matrix_linear_combination(V, domain):
+    """Test linear combinations involving ufl.Matrix."""
+    # Matrix requires a row space and column space
+    A = ufl.Matrix(V, V)
+    c = ufl.Constant(domain)
+    expr = 2.0 * A + (0.3 + c**2) * A
+
+    res = extract_linear_combination(expr)
+
+    assert len(res) == 2
+    # NOTE: Matrices store scalar weights under `weights`
+    assert (2.0, A) in res
+    assert (0.3 + c**2, A) in res
+
+
+def test_cofunction_linear_combination(V):
+    """Test linear combinations involving ufl.Cofunction."""
+    # Cofunction requires a dual space (it will raise an error if given a primal space)
+    V_dual = V.dual()
+    c = ufl.Cofunction(V_dual)
+
+    # NOTE: Cofunctions can only be multiplied by integers
+    expr = -2 * c + 5 * c
+
+    res = extract_linear_combination(expr)
+
+    assert len(res) == 2
+    assert (-2.0, c) in res
+    assert (5.0, c) in res
+
+
+def test_matrix_nonlinear_error(V):
+    """Test that matrices are protected by the same non-linear guardrails."""
+    A = ufl.Matrix(V, V)
+    u = ufl.Coefficient(V)
+
+    with pytest.raises(
+        ValueError, match=r"Non-linear expression detected: product of two spatial functions."
+    ):
+        # Cannot multiply a matrix by a coefficient algebraically in this block
+        extract_linear_combination(A * u)
--- a/test/test_extract_linear_combination.py
+++ b/test/test_extract_linear_combination.py
@@ -167,14 +167,12 @@ def test_cofunction_linear_combination(V):
     # Cofunction requires a dual space (it will raise an error if given a primal space)
     V_dual = V.dual()
     c = ufl.Cofunction(V_dual)
-
-    # NOTE: Cofunctions can only be multiplied by integers
-    expr = -2 * c + 5 * c
-
+    d = ufl.Constant(V.ufl_domain())
+    expr = -4 * d * c + 5.0 * c
     res = extract_linear_combination(expr)
 
     assert len(res) == 2
-    assert (-2.0, c) in res
+    assert (-4 * d, c) in res
     assert (5.0, c) in res
 
 
--- a/test/test_extract_linear_combination.py
+++ b/test/test_extract_linear_combination.py
@@ -1,0 +1,141 @@
+import pytest
+from utils import LagrangeElement
+
+import ufl
+from ufl.algorithms.extract_linear_combination import extract_linear_combination
+
+
+@pytest.fixture
+def domain():
+    return ufl.Mesh(LagrangeElement(ufl.triangle, 1, (2,)))
+
+
+@pytest.fixture
+def V(domain):
+    """Standard spatial function space."""
+    el = LagrangeElement(domain.ufl_cell(), 3)
+    return ufl.FunctionSpace(domain, el)
+
+
+@pytest.fixture
+def V_vec(domain):
+    """Vector function space for indexed tests."""
+    return ufl.FunctionSpace(domain, LagrangeElement(domain.ufl_cell(), 2, (3,)))
+
+
+def test_valid_linear_combinations(V):
+    """Test standard valid combinations.
+
+    Sum, Product, Division, Negative, and float/int terminals.
+    """
+    u = ufl.Coefficient(V)
+    v = ufl.Coefficient(V)
+
+    # Includes: IntValue (3), FloatValue (1.5), Sum, Negative (-v/4.0), Product, Division
+    expr = 2.0 * u - v / 4.0 + ufl.as_ufl(3) * u - ufl.classes.FloatValue(1.5) * v
+
+    res = extract_linear_combination(expr)
+
+    assert len(res) == 4
+    # The DAG traversal evaluates leaves to roots, maintaining term extraction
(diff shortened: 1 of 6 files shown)


test/test_extract_linear_combination.py
```diff
@@ -0,0 +1,184 @@
+import pytest
```
jorgensd, line 7:
```suggestion
# Copyright (C) 2026 Jørgen S. Dokken
#
# This file is part of UFL (https://www.fenicsproject.org)
#
# SPDX-License-Identifier:    LGPL-3.0-or-later

import pytest
```
Add copyright header.

test/test_extract_linear_combination.py
```diff
@@ -0,0 +1,190 @@
...
+    assert (0.3 + c**2, A) in res
+
+
+def test_cofunction_linear_combination(V):
+    """Test linear combinations involving ufl.Cofunction."""
+    # Cofunction requires a dual space (it will raise an error if given a primal space)
+    V_dual = V.dual()
+    c = ufl.Cofunction(V_dual)
+
+    # NOTE: Cofunctions can only be multiplied by integers
```
pbrubeck: integers??

ufl/algorithms/extract_linear_combination.py
```diff
@@ -0,0 +1,204 @@
...
+    @process.register(ufl.Matrix)
+    def _(self, o, **kwargs):
+        # Cofunctions and Matrices are symbolic algebraic terminals.
+        # They act purely as spatial fields in linear combinations.
+        return [(ufl.as_ufl(1.0), o)]
+
+
+def extract_linear_combination(
+    expr: ufl.core.expr.Expr,
+) -> list[tuple[ufl.core.expr.Expr, ufl.classes.Coefficient]]:
```
pbrubeck: CoefficientBase?

ufl/algorithms/extract_linear_combination.py
```diff
@@ -0,0 +1,204 @@
...
+        raise NotImplementedError(
+            "Direct array assignment of indexed vector components is not supported."
+        )
+
+    @process.register(ufl.Cofunction)
+    @process.register(ufl.Matrix)
+    def _(self, o, **kwargs):
+        # Cofunctions and Matrices are symbolic algebraic terminals.
+        # They act purely as spatial fields in linear combinations.
+        return [(ufl.as_ufl(1.0), o)]
```
pbrubeck: These can be merged into the Coefficient register

ufl/algorithms/extract_linear_combination.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def extract_linear_combination(
```
pbrubeck, line 186: Can this unpack a `FormSum` into the list of tuples?

test/test_extract_linear_combination.py
jorgensd: Never mind, I had an error at some point that lead me to believe this, but it is not true.

ufl/algorithms/extract_linear_combination.py
jorgensd: Yes, that is what the `test_cofunction_linear_combination` test checks.

test/test_extract_linear_combination.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def test_ufl_zero_node(V):
+    """Test that UFL Zero nodes are handled correctly (usually evaluates to an empty list)."""
+    u = ufl.Coefficient(V)
+
+    # Create an explicit Zero node with the shape of u
+    z = ufl.classes.Zero(u.ufl_shape, u.ufl_free_indices, u.ufl_index_dimensions)
+
+    expr = u + z
```
pbrubeck: Aren't we eagerly simplifying u+0->u?

pbrubeck:
Wouldn't `dolfinx.fem.Function` inherit from `ufl.Coefficient`? Why do you need to subclass the DAGTraverser in the example from the description?

jorgensd:
> Wouldn't `dolfinx.fem.Function` inherit from `ufl.Coefficient`? Why do you need to subclass the DAGTraverser in the example from the description?

I would need to sub-class it to detect a `Real-valued` finite element, which acts as a differentiable constant, and only has DOLFINx specific data, such as `ufl_element().is_real` which is not part of the UFL standard.

test/test_extract_linear_combination.py
jorgensd: You are right, so this test just checks that one can extract a coefficient form itself. Will remove it.

pbrubeck:
> I would need to sub-class it to detect a `Real-valued` finite element, which acts as a differentiable constant, and only has DOLFINx specific data, such as `ufl_element().is_real` which is not part of the UFL standard.

But the UFL standard can identify piecewise and global constants. I think we should try to do that here, since Firedrake also has a Real element

jorgensd:
> > I would need to sub-class it to detect a `Real-valued` finite element, which acts as a differentiable constant, and only has DOLFINx specific data, such as `ufl_element().is_real` which is not part of the UFL standard.
> 
> But the UFL standard can identify piecewise and global constants. I think we should try to do that here, since Firedrake also has a Real element

Great point! I've adjusted the code, and it passes my test suite in DOLFINx with real element coefficients.

ufl/algorithms/extract_linear_combination.py
```diff
@@ -0,0 +1,203 @@
...
+        super().__init__(**kwargs)
+
+    @singledispatchmethod
+    def process(self, o: ufl.classes.Expr, **kwargs):
+        """Fallback for any unsupported node types."""
+        raise ValueError(f"Unsupported UFL node type for linear combinations: {type(o)}")
+
+    @process.register(ufl.coefficient.BaseCoefficient)
+    def _(self, o, **kwargs):
+        raise NotImplementedError(f"Unsupported UFL node type for linear combinations: {type(o)}")
```
pbrubeck: Why aren't we supporting this case?

jorgensd: In the same way that we are not supporting `ufl.core.expr.Expr` as a general term. 
The  only two classes that for now uses `BaseCoefficient` is `Cofunction` and `Function`, which have specializations for in the code. This would be a good check if someone implements a new class based on BaseCoefficient, which they would have to check if the current implementation is correct.

pbrubeck: Wouldn't we want linear combinations to just work on newly implemented subclasses of BaseCoefficient?

jorgensd: I guess it depends on how it should be handled.
Currently there is a split between how Coefficient and CoFunction is handled, as coefficient checks for 
`ufl.checks.is_scalar_constant_expression(o)` to factor it out of a linear combination, while this is not the case for the CoFunction, as it is always  a scalar_constant_expression.

pbrubeck: I don't see why would one want distinct treatment for Coefficient, Cofunction and Matrix.

ufl/algorithms/extract_linear_combination.py
```diff
@@ -0,0 +1,209 @@
...
+    @process.register(ufl.Cofunction)
+    @process.register(ufl.Matrix)
+    def _(self, o, **kwargs):
+        return [(ufl.as_ufl(1.0), o)]
+
+    @process.register(ufl.classes.Coefficient)
+    def _(self, o, **kwargs):
+        # Check for real-valued elements
+        if ufl.checks.is_scalar_constant_expression(o):
+            return o
```
pbrubeck: What are the assumptions here? Are we allowing affine expressions, i.e. `sum_i a_i v_i + Constant`?

jorgensd: The assumption here is that if we have (sum a_i v_i), then in some cases a_i might be a scalar valued real space coefficient, and should therefore be returned as a float rather than the tuple (1, a_i).

We are not allowing for a_i v_i + constant in this PR.

pbrubeck: Why is this unique to Coefficient? We should give no special treatment to Coefficient vs Cofunction

PR #507, opened 2026-08-24
Further extensions to extract linear combination 

jorgensd:
operations required for irksome + further improvements suggested by pablo.

Created with Gemini (August 2026), but tested and verified by me.


ufl/algorithms/extract_linear_combination.py
```diff
@@ -173,14 +174,57 @@ def _(self, o, *operands, **kwargs):
...
     @process.register(ufl.classes.Indexed)
+    @DAGTraverser.postorder
+    def _(self, o, *operands, **kwargs):
+        base_res, indices_res = operands
+
+        # We ONLY allow UFL's free indices. Fixed components (FixedIndex)
+        # or slices (SliceIndex) do not represent the whole operator.
+        if not all(isinstance(idx, ufl.classes.Index) for idx in indices_res.indices()):
+            raise ValueError(
+                "Explicit component indexing (e.g., u[0]) is not supported. "
```
pbrubeck: Why are we not supporting `FixedIndex`?

jorgensd: I've now extended the code to cover FixedIndex and slicing @pbrubeck. See supporting tests.

ufl/algorithms/extract_linear_combination.py
```diff
@@ -173,14 +174,43 @@ def _(self, o, *operands, **kwargs):
...
     @process.register(ufl.classes.ComponentTensor)
-    def _(self, o, **kwargs):
-        raise NotImplementedError(
-            "Direct array assignment of indexed vector components is not supported."
-        )
+    @DAGTraverser.postorder
+    def _(self, o, *operands, **kwargs):
+        expr_res, indices_res = operands
+
+        if isinstance(expr_res, list):
```
pbrubeck, line 199: Is this always a list? Why is a list stored inside a UFL node? That makes it un-hashable. Shouldn't we have tuplify it?

jorgensd: The return argument from the dag traverser is either a single float or a list of tuples [(float_expr, operand),....].
This dag-traverser is a bit special, in the sense that it doesn't convert a ufl expression into a new expression, but decomposes it into the (float_expr, operand) tuples.

pbrubeck: Alright, this makes everything clearer. I thought that operands we the original ones for the node being processed, but they are the post-processed ones.

pbrubeck (approved):

PR #508, opened 2026-08-26
Improve/fix adjoint computation on mixed function space

jorgensd:
When I introduced this for mixed function spaces (https://github.com/FEniCS/ufl/pull/352), I made a mistake in the test that masked an issue,
Argument.part() should not swap.

New improved tests and cleaner logic.


ufl/algorithms/formtransformations.py
```diff
@@ -490,75 +490,89 @@ def compute_energy_norm(form, coefficient):
...
+        for i, row in enumerate(form_blocked):
+            for j, block in enumerate(row):
+                if block is not None:
+                    v, u = block.arguments()
+                    if reordered_arguments is not None:
+                        new_v = reordered_arguments[i][1]
+                        new_u = reordered_arguments[j][0]
+                    else:
+                        new_v = Argument(v.ufl_function_space(), number=u.number(), part=v.part())
+                        new_u = Argument(u.ufl_function_space(), number=v.number(), part=u.part())
```
pbrubeck, line 576: Could you add a comment here saying that we only need to swap the number? I think it'd be convenint to introduce `Argument.reconstruct()`. In this way Firedrake can stop subclassing Argument.

pbrubeck (changes requested):

ufl/algorithms/formtransformations.py
jorgensd: Note added (it is part of describing the function space).
I will build a PR on top of this introducing reconstruct.

PR #511, opened 2026-09-08
Give each pull back an inverse

pbrubeck:
# Description

A pull back maps a function on the reference cell to the physical cell. UFL had no map in the other direction, so a form compiler that has to evaluate a physical expression on the reference cell — to dual-evaluate an `Interpolate`, for instance — carried its own copy of the Piola maps. TSFC has one.

This adds `apply_inverse` to `AbstractPullback` and implements it for every pull back that has an inverse:

| pull back | `apply` | `apply_inverse` |
|---|---|---|
| `IdentityPullback` | `g` | `g` |
| `ContravariantPiola` | `(1/detJ) J g` | `detJ K g` |
| `CovariantPiola` | `Kᵀ g` | `Jᵀ g` |
| `L2Piola` | `g / detJ` | `g detJ` |
| `DoubleContravariantPiola` | `(1/detJ)² J g Jᵀ` | `detJ² K g Kᵀ` |
| `DoubleCovariantPiola` | `Kᵀ g K` | `Jᵀ g J` |
| `CovariantContravariantPiola` | `(1/detJ) Kᵀ g Jᵀ` | `detJ Jᵀ g Kᵀ` |
| `MixedPullback` | per sub-element | per sub-element |
| `SymmetricPullback` | per unique block | per unique block |

`PhysicalPullback` and `CustomPullback` are their own inverse. `UndefinedPullback` inherits the `NonStandardPullbackException` from `AbstractPullback`, as it does for `apply`.

The test asserts the property that matters rather than the formulas: pushing a reference function forward and mapping it back returns the original, once `remove_component_tensors` and `cancel_jacobian_products` have cancelled the Jacobian products the round trip leaves behind. That closes exactly for eight of the nine, mixed and symmetric elements included. The `L2Piola` round trip leaves `detJ / detJ` standing, because cancelling a scalar factor is not something `cancel_jacobian_products` does, so that one is asserted directly instead.

## Stacking

`pbrubeck/pullback-apply-inverse` sits on `pbrubeck/interpolate-holes` (FEniCS/ufl#497), so that it pairs with `pbrubeck/form-interp-tsfc` in Firedrake. That one branch is also the head of FEniCS/ufl#511, which proposes the same commit upstream; #497 is in that diff for the same reason.

The pull back work itself does not depend on #497 and can be read on its own.

Nothing calls `apply_inverse` yet. #69 uses it to lower interpolation in `compute_form_data`, which lets TSFC delete its copy of the maps.

https://claude.ai/code/session_0155wQMTMSGiF9a6TbyTVK5P

### DIFF
--- a/test/test_apply_coefficent_split.py
+++ b/test/test_apply_coefficent_split.py
@@ -12,6 +12,7 @@ from ufl.algorithms.apply_coefficient_split import apply_coefficient_split
 from ufl.classes import (
     ComponentTensor,
     Indexed,
+    Interpolate,
     ListTensor,
     PositiveRestricted,
     ReferenceGrad,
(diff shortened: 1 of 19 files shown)


jorgensd:
Im positive to this change as I also had to handroll the inverses as some point;) will try to review on Thursday

jorgensd (approved):

PR #513, opened 2026-09-08
Fix Grad/SpatialCoordinate pullback for coefficients on a codimension-1/2 companion (submesh) domain

jorgensd:
## Problem
A `Coefficient` (or `Argument`) defined on a domain that is a facet or ridge of the actual integration domain (in DOLFINx a submesh of codim 1/2), integrated via a `ds/dr(domain=parent_mesh)` measure got the wrong value for `Grad/Dx` and `SpatialCoordinate`.

## Cause
`GeometryLoweringApplier` always lowered a domain's `Jacobian`/`SpatialCoordinate` purely in terms of that domain's own coordinate field, with no way to express it via the integration domain's `FacetJacobian`/`RidgeJacobian` instead.

FFCx has no separate coordinate buffer for such a companion domain — only for the integration domain — so the lowered expression was unusable in general (in FFCx's case, it silently read the wrong vertices out of the parent's buffer, giving a value independent of which facet was actually being integrated). There is no point of sending in a separate coordinate buffer for a sub-mesh, as it is just the restriction of coordinates of the original cell.

## Fix and firedrake compatibility

Jacobians on companion domains (submeshes) are now rewritten via the integration domain's `FacetJacobian`/`RidgeJacobian` , and a companion domain's SpatialCoordinate is rewritten as the integration domain's own `SpatialCoordinate`. Same-topological-dimension companion domains (codim-0 submeshes) and domains declared via `Measure(..., intersect_measures=...)` (@pbrubeck) are left untouched, since both already have working, independent handling.

## Additional error handling

Mismatched integral types and geometric-dimension mismatches now raise a clear NotImplementedError/ ValueError instead of silently lowering to something meaningless.

## Verification
- New tests added to check the substitution of said quantities (that fails on main).
- `pytest test/ (full UFL suite, 1001 passed)`
 

This PR has been created by me with the assistance of CLAUDE (Sonnet 5). I have reviewed the code and take repressibility for the correctness.


test/test_apply_geometry_lowering.py
```diff
@@ -0,0 +1,160 @@
+"""Tests for geometry lowering of a coefficient defined on a companion domain.
+
+A companion domain is a domain distinct from the integration domain that
+names one of its facets/ridges (e.g. the domain of a coefficient defined
+on a submesh, integrated via a measure on the parent mesh).
+It's Jacobian/SpatialCoordinate is rewritten in the terms of the integration domain's
```
pbrubeck, line 6:
```suggestion
Its Jacobian/SpatialCoordinate is rewritten in the terms of the integration domain's
```

test/test_apply_geometry_lowering.py
```diff
@@ -0,0 +1,160 @@
...
+    codomain = Mesh(LagrangeElement(interval, 1, (2,)))
+    space = FunctionSpace(codomain, LagrangeElement(interval, 1))
+    u = Coefficient(space)
+    ds = Measure("ds", domain=domain)
+
+    integrand = _lower(Dx(u, 0) * ds)
+
+    for cls in (Jacobian, JacobianInverse):
+        for obj in extract_type(integrand, cls):
+            assert extract_unique_domain(obj) != codomain
```
pbrubeck, line 71: Isn't this a stronger assertion?
```suggestion
            assert extract_unique_domain(obj) == domain
```

jorgensd:
Going to close this as it introduced a massive headache in FFCx due to dof permutation. Will open a new PR that only touches FFCx and keeps UFL as is.

PR #515, opened 2026-09-12
Interpolate

schnellerhase:


ufl/core/interpolate.py
```diff
@@ -155,6 +158,15 @@ def signature(slot):
...
         return self._signature
 
+    def target_space(self) -> AbstractFunctionSpace:
+        """Return the function space that the expression is interpolated into.
+
+        Unlike `ufl_function_space` this does not depend on the argument
+        numbering, so it stays correct when the interpolated expression holds an
+        argument of the enclosing form.
+        """
+        return self._function_space
```
pbrubeck, line 168: What does this mean for the adjoint, i.e. `Interpolate(Argument, Cofunction)`?

ufl/core/interpolate.py
```diff
@@ -66,7 +68,8 @@ def __init__(self, expr, v):
         # Check valid argument numbering
         if expr_arg_numbers & dual_arg_numbers:
             raise ValueError("Same argument numbers in first and second operands to interpolate.")
-        if expr_arg_numbers | dual_arg_numbers not in [set(), {0}, {0, 1}]:
+        argument_numbers = expr_arg_numbers | dual_arg_numbers
+        if argument_numbers and argument_numbers != set(range(len(argument_numbers))):
```
pbrubeck, line 72: why do we need this check? I think the main purpose is to ensure unique argument numbers, but we should not be constraining the arguments to always start from 0 and be contiguous.

ufl/core/interpolate.py
```diff
@@ -55,7 +55,9 @@ def __init__(self, expr, v):
         if isinstance(v, AbstractFunctionSpace):
             if is_dual(v):
                 raise ValueError("Expecting a primal function space.")
-            n = 1 if expr_arg_numbers == {0} else 0
+            # Take the lowest free number, so that an expression linear in
+            # several arguments stays representable.
+            n = min(set(range(len(expr_arg_numbers) + 1)) - expr_arg_numbers)
```
pbrubeck, line 60: We should add tests for this

ufl/algorithms/balancing.py
```diff
@@ -32,11 +33,21 @@
...
 
 
+def _is_modifiable(expr):
+    """Check whether modifiers may be applied to ``expr``.
+
+    An interpolation is modifiable like a terminal: it is a finite element field
+    of its target space, and the expression it interpolates is evaluated
+    separately.
+    """
+    return expr._ufl_is_terminal_ or isinstance(expr, Interpolate)
```
pbrubeck, line 43: this pattern repeats through other files. I don't agree with the name. In which sense are terminals "modifiable"?

ufl/algorithms/apply_derivatives.py
```diff
@@ -1996,18 +2034,26 @@ def __iadd__(self, other):
...
     """Apply derivatives to an expression.
 
     Args:
         expression: A Form, an Expr or a BaseFormOperator to be differentiated
+        differentiate_through_operators: If True, differentiate a base form
+            operator by differentiating its operand, which is exact because the
+            operator is linear in it, instead of recording the derivative as an
+            action to be assembled separately. This composes to higher
+            derivatives and keeps the result an ordinary form, which a form
+            compiler that evaluates the operator cell-locally needs.
```
pbrubeck, line 2047: Is this needed to enable `derivative(interpolate(u, V1)*dx, u)`? Should this be the default?

PR #517, opened 2026-09-20
Fix complex arity checks for linear operators

pbrubeck:
## Summary

UFL previously treated several compound linear differential operators as nonlinear during direct complex arity checking, rejecting valid mixed forms such as mixed Poisson formulations.

This change:

- Migrates `ArityChecker` from `MultiFunction` to `DAGTraverser`.
- Registers all linear differential operators, including div, curl, nabla, and reference variants.
- Preserves nonlinear-operator cutoffs, interpolation handling, and existing arity rules.
- Records arity dependencies as `(argument, is_conjugated)` pairs for complex-mode validation.
- Fixes the comparison checker so indexing metadata is treated as real-valued. This prevents expressions produced by algebra lowering, such as `IndexSum` and `ComponentTensor`, from being rejected as spuriously complex while preserving errors for genuinely complex comparisons.
- Adds regression coverage for complex arity and comparison checking, including lowered real expressions.

## Verification

- 1034 tests passed
- Ruff passed
- Mypy passed


ufl/algorithms/check_arities.py
```diff
@@ -189,10 +262,15 @@ def list_tensor(self, o, *ops):
 
 
 def check_integrand_arity(expr, arguments, complex_mode=False):
-    """Check the arity of an integrand."""
+    """Check the arity of an integrand.
+
+    The arity extraction records each argument as an
+    ``(argument, is_conjugated)`` pair so complex-mode checks can validate the
+    required conjugation of test and trial functions.
```
pbrubeck:
```suggestion
    required conjugation state of test and trial functions.
```

pbrubeck: or should we say "unconjugation of trial functions"?

jorgensd:
```suggestion
    Arity extraction records each argument as an ``(argument, is_conjugated)`` pair. This enables
    complex-mode checks to validate the required state: test functions are always conjugated, whereas
    trial functions remain unconjugated.
```

jorgensd (approved):

ufl/algorithms/check_arities.py
```diff
@@ -189,10 +262,15 @@ def list_tensor(self, o, *ops):
 
 
 def check_integrand_arity(expr, arguments, complex_mode=False):
-    """Check the arity of an integrand."""
+    """Check the arity of an integrand.
+
+    Arity extraction records each argument as an ``(argument, is_conjugated)`` pair. This enables
+    complex-mode checks to validate the required state: test functions are always conjugated, whereas
+    trial functions remain unconjugated.
```
jorgensd:
```suggestion
    complex-mode checks to validate the required state: test functions are always conjugated,
    whereas trial functions remain unconjugated.
```

